Issue #3, opened 2016-05-15
SSL error with OpenSSL


nloveladyallen:
Rather ironically, livechecking OpenSSL returns an SSL error:

    Error: SSL_connect SYSCALL returned=5 errno=0 state=SSLv2/v3 read server hello A

youtux:
This issue seems to be caused by the fact that Mac OS ships with an old version of OpenSSL:
```
$ openssl version
OpenSSL 0.9.8zh 14 Jan 2016
```
which does not seem to support TLSv1 nor TLSv2 (source: http://stackoverflow.com/a/33295040).
Unfortunately Homebrew uses the version of Ruby that is shipped with Mac OS, which is compiled against this version of OpenSSL.

I am not aware about how to workaround this issue.
If anyone has any suggestion, it will be much appreciated.

nloveladyallen:
Could you scrape the latest version from Releases on their [GitHub repo](https://github.com/openssl/openssl)?

nloveladyallen:
Or their [ftp server](ftp://ftp.openssl.org/source/)

PR #4, opened 2016-05-19
Fix openssl livecheck: switch to github repo

nloveladyallen:
Fix to #3


youtux:
Thanks for the contribution, but I'd rather prefer to rely on git rather than HTML pages.
Sorry for the delay of the fix.

PR #7, opened 2016-05-24
Make SourceForge regex case insensitive

nloveladyallen:
Otherwise any SourceForge project with uppercase letters in its name fails to check.


youtux:
👍

PR #9, opened 2016-05-30
Add sub-command functionality

nloveladyallen:
Dead simple sub-command implementation. Each sub-command is a script (in the sense of "no classes, no methods, no nothing") in livecheck/commands. It would definitely be more "correct" to do it with classes, so feel free to close this if you care about that kind of thing.

### DIFF
--- a/cmd/brew-livecheck.rb
+++ b/cmd/brew-livecheck.rb
@@ -94,7 +94,7 @@ if ARGV.flag?("--help")
   exit 0
 end
 
-require? "livecheck/commands/" + ARGV[0]
+require? "livecheck/commands/" + ARGV.named[0]
 
 formulae_to_check =
   case
--- a/cmd/brew-livecheck.rb
+++ b/cmd/brew-livecheck.rb
@@ -29,6 +29,15 @@ usage = <<-EOF.undent
   -d, --debug       show debugging info
 EOF
 
+# Taken directly from Homebrew
+def require?(path)
+  require path
+rescue LoadError => e
+  # HACK: ( because we should raise on syntax errors but
+  # not if the file doesn't exist. TODO make robust!
+  raise unless e.to_s.include? path
+end
+
 if (Pathname.new(File.expand_path("..", __FILE__)).basename).to_s == "bin"
   opoo <<-EOS.undent
     It seems you are using an old version of homebrew-livecheck.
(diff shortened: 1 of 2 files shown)


livecheck/commands/cat.rb
```diff
@@ -0,0 +1,4 @@
+# Ignores all formulae but first (this is by design)
+livecheckable_path = LIVECHECKABLES_PATH / "#{ARGV[1]}.rb"
+puts File.open(livecheckable_path).read
+exit 0
```
youtux: is it necessary?

cmd/brew-livecheck.rb
```diff
@@ -85,7 +94,9 @@ def print_latest_version(formula)
   exit 0
 end
 
-formulae_to_check = 
+require? "livecheck/commands/" + ARGV[0]
```
youtux: use `ARGV.named[0]`, so that we can still pass flags like "-d"

livecheck/commands/cat.rb
```diff
@@ -0,0 +1,4 @@
+# Ignores all formulae but first (this is by design)
+livecheckable_path = LIVECHECKABLES_PATH / "#{ARGV[1]}.rb"
```
youtux: use ARGV.named[1]

cmd/brew-livecheck.rb
nloveladyallen: Fixed with 64b0499.

livecheck/commands/cat.rb
nloveladyallen: Fixed with 64b0499.

livecheck/commands/cat.rb
```diff
@@ -0,0 +1,4 @@
+# Ignores all formulae but first (this is by design)
+livecheckable_path = LIVECHECKABLES_PATH / "#{ARGV.named[1]}.rb"
+puts File.open(livecheckable_path).read
+exit 0
```
youtux: instead of `exit 0`, can you check if the command exists, and in that case execute it and exit in the main `cmd/brew-livecheck.rb` file?

nloveladyallen: As in `brew livecheck cat edit formula`? (once `edit` is implemented)

youtux: Actually what I meant is that I don't want to write `exit 0` in every subcommand, it should be implicit. In order to do this, we should check in `cmd/brew-livecheck.rb` if the command (`ARGV.named[0]`) exists under the `livecheck/commands/` path, and if it does then execute it and exit. The `exit 0` should be performed in `cmd/brew-livecheck.rb`

nloveladyallen: Fixed with 745a586.

youtux:
Thank you! 👍

PR #22, opened 2017-07-14
more livechecks 3

ilovezfs:


ilovezfs:
@youtux when merging these, we should `rebase and merge` not `squash and merge`. I'm not sure if I can fix it now but we'll see.

youtux:
Why is that?

ilovezfs:
To preserve one commit per formula.

youtux:
I see. We didn't really specify a workflow for this project, maybe it is better to document this.

PR #28, opened 2017-09-03
Fixed "redirection forbidden" errors for pyqt and sip

aluttik:


ilovezfs:
@aluttik please make this one commit per formula :)

aluttik:
@ilovezfs mb, I'll get on that later today

aluttik:
@ilovezfs: Sorry that took so long, I totally forgot about this. Is that better?

ilovezfs:
Thanks @aluttik! Shipped :)

PR #95, opened 2018-08-26
Generalize npm livechecks

tommyang:
This PR:
* adds euristic for npm
* removes most of the livechecks with `npmjs.com` urls since they are now covered by this euristic
  * `bower.rb` & `imageoptim-cli.rb` are not removed. The reason is that their corresponding `homebrew-core` formula urls are not under `registry.npmjs.org`. 

Not sure if I am doing this right or not since this is my first non-formula PR for Homebrew.


commitay:
Thanks @tommyang!

PR #96, opened 2018-09-13
libx11: add livecheck

maxim-belkin:
**Tap:** linuxbrew/xorg


maxim-belkin:
it might be better to implement this in ~euristics.rb~ euristic.rb

maxim-belkin:
superseded by Homebrew/homebrew-livecheck#98

PR #97, opened 2018-09-13
libxau: add livecheck

maxim-belkin:
**Tap:** linuxbrew/xorg


maxim-belkin:
Closing this one while I'm trying to come up with a generic solution

PR #98, opened 2018-09-13
euristic.rb: heuristics bits for Linuxbrew/xorg tap

maxim-belkin:
Livecheck heuristics for files hosted on x.org and freedesktop.org websites (used in Linuxbrew/xorg tap)

### DIFF
--- a/livecheck/euristic.rb
+++ b/livecheck/euristic.rb
@@ -115,6 +115,14 @@ def version_euristic(urls, regex = nil)
         version = Version.new(match)
         match_version_map[match] = version
       end
+    when url =~ %r{(...\.x\.org|freedesktop.org)}
+      package = url.split("/")[-1].split("-")[0..-2].join("-")
+      page_url = url.split("/")[0..-2].join("/")
+      regex = /#{package}-([0-9.]+(?:\.[0-9.]+)*)\.tar\.[a-zA-Z0-9]*/
+      page_matches(page_url, regex).each do |match|
+        version = Version.new(match)
+        match_version_map[match] = version
+      end
     when regex
       # Fallback
       page_matches(url, regex).each do |match|


maxim-belkin:
With the current single-repo + non-extendable approach, all heuristics have to live in the file that is modified in this PR. If one could add their own set of heuristics in a directory in their tap - that would've been a much better (scalable and maintainable) approach.

- There is no "Linuxbrew heuristic".
- In Linuxbrew/xorg there are 176 formulae most of which share the same "heuristic" so adding livechecks manually to every one would be (a) painful (b) violate DRY.

MikeMcQuaid:
Personally I've had no involvement in this repo so ultimately I defer to you @commitay. My personal preference would be to merge this if as we may end up wanting to use these for future formulae and it doesn't seem to do any harm.

> I think adding support for `livechecks` in formulae would be a better way of supporting Linuxbrew (and other taps) rather than mixing them into this repo.

I do agree with this, though. I'd like to see a `livecheck` DSL for this which allows both custom `livecheck` in formulae and those that are automatically inferred or manually tagged (e.g. `livecheck :github_release`, `livecheck :xorg`) based on a URL and some of the underlying logic live in Homebrew/brew.

youtux:
When I originally wrote the heuristic.rb file, the intention was to try and fetch versions only using some of the "well known" strategies: tags from git repositories, versions from npmjs registry, potentially versions from pypi, etc. The reason for this is that I didn't want to write thousands of formulae.

Later on the module grew with more and more euristics for specific websites, like gnu.org and sourceforge.
Eventually it got more and more logic to handle specific use cases, and at that time I should have been more active in the project and try to push for a better way.

Ideally I would like to export the strategies in a separate file and get rid of the euristics module and the Livecheckables folder, and let each Formula define its way to fetch the latest version. But this would require this project to be merged in the core of Homebrew, and even if everybody agreed on that, it would require quite some effort.

About this case, I'd reather see 176 new Livecheckables merged than yet another patch to the euristic.rb (aside from renaming it in heuristics.rb).

MikeMcQuaid:
> But this would require this project to be merged in the core of Homebrew, and even if everybody agreed on that, it would require quite some effort.

To be clear: from my perspective I very much agree with that 👍

maxim-belkin:
> About this case, I'd reather see 176 new Livecheckables merged than yet another patch to the euristic.rb (aside from renaming it in heuristics.rb).

You're kidding, right?

MikeMcQuaid:
@maxim-belkin That comment is unnecessarily rude. If you wish to make a technical argument: make one. If you wish to vent: do so to `/dev/null`.

maxim-belkin:
Wait, I'm confused. He was not kidding? wow! OK, so here is my response...

1. I was (and still am) genuinely surprised to see a suggestion do something hundreds of times (and receive 176 new files) instead of merging one piece of code that does the same thing "in one go" and is only few-lines long! Before your, @MikeMcQuaid, comment I had I a feeling that @youtux was just trolling me: at first he said  _"I didn't want to write thousands of formulae"_ and then he suggested me to do (almost exactly) that: _"I'd reather see 176 new Livecheckables merged than yet another patch to the euristic.rb"_.  But it looks like you guys were serious... well, I'm confused (and surprised, I guess). Why exactly would 176 separate livecheckables be better?

2. There was an argument that this repo should not serve any other tap but homebrew-core. This is a decision that Homebrew community / leadership team / lead maintainer  has to make. It would be sad if such a decision is made but it has to be explicitly stated so that other maintainers (like myself) don't waste their time trying to extend this repo to support theirs. 

Other thoughts:

3. I don't think **this** (Homebrew/homebrew-livecheck) repo is the right place to hold individual livecheckables for Linuxbrew/xorg (and any other tap, actually, including homebrew-core for a number of reasons but that's a different topic). I'll just note that it took 4 days to receive even an initial response to the PR.

4. A scalable solution IMHO would be to allow taps extend `heuristics` and/or create their own Livecheckables directories. (I could work on that part of things but not sure you guys share this view so don't want to waste my time.) Individual Livecheckable files, in my opinion, should be used as a last resort in exceptional cases when heuristics does not work.

5. I'll be fine if this PR is not merged (I'll just continue to use my "hacked" version of heuristics file). I'm just surprised to see resistance to a contribution that extends capabilities of the tool. Why?

youtux:
My last comment was extreme on purpose, I just wanted to make a point about the `euristics.rb` file becoming a giant piece of unmaintainable code.

I think that it would be very nice for this project to be merged into homebrew, but I cannot allocate much time for that, unfortunately.

@maxim-belkin if you can try to implement what you describe in your 4th point I will surely appreciate that. Even better, if homebrew maintainers agree to have the livecheck command and the related changes to the DSL implemented in the core, and if you want to try to do that, I would be even happier.

PR #99, opened 2018-09-16
libsoup: update regex

jeffreywildman:


commitay:
Thanks @jeffreywildman!

PR #101, opened 2018-12-03
bash-completion: add livecheck

Moisan:


Livecheckables/bash-completion.rb
```diff
@@ -0,0 +1,4 @@
+class BashCompletion
+  livecheck :url => "https://src.fedoraproject.org/repo/pkgs/bash-completion/",
```
youtux: We should not check against `fedoraproject.org`. We should use the official repository: `https://github.com/scop/bash-completion`.

Moisan:
Closing since `bash-completion` version is fixed, `bash-completion@2` is the latest version.

PR #104, opened 2018-12-24
rpm: add livecheck

Moisan:


youtux:
I would prefer using the approach of https://github.com/Homebrew/homebrew-livecheck/blob/f3b022a30b19332c95ac6ac3cb469a828710ac49/Livecheckables/boost-build.rb: using the `https://github.com/rpm-software-management/rpm.git` url with a regex like `/rpm-([0-9\.]+)-release/`

PR #106, opened 2019-01-26
Add livecheckables

jeffreywildman:

PR #109, opened 2019-02-17
Add livecheckables

jeffreywildman:
Effectively re-opening #106 without GNOME-related packages.


youtux:
Thanks @jeffreywildman

PR #111, opened 2019-02-17
Remove individual GNOME livecheckables in favor of generalized check

jeffreywildman:
Test before/after with:

```bash
brew livecheck at-spi2-atk at-spi2-core glib gnome-latex gnumeric gobject-introspection goffice gom goocanvas gsettings-desktop-schemas gtk-vnc gtksourceview@4 jsonrpc-glib libdazzle librsvg libsoup template-glib tepl
```

### DIFF
--- a/Livecheckables/goffice.rb
+++ b/Livecheckables/goffice.rb
@@ -1,4 +1,3 @@
 class Goffice
-  livecheck :url => "https://download.gnome.org/sources/goffice/0.10/",
-            :regex => /href="LATEST-IS-([0-9\.]+)"/
+  livecheck :url => "https://download.gnome.org/sources/goffice/cache.json"
 end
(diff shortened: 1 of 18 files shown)


Livecheckables/goffice.rb
```diff
@@ -1,4 +1,3 @@
 class Goffice
-  livecheck :url => "https://download.gnome.org/sources/goffice/0.10/",
-            :regex => /href="LATEST-IS-([0-9\.]+)"/
+  livecheck :url => "https://download.gnome.org/sources/goffice/cache.json"
```
youtux: can we remove this file as well?

jeffreywildman: I tried, but livecheck ends up using the formula's head block [url](https://github.com/Homebrew/homebrew-core/blob/master/Formula/goffice.rb#L14) from github.com instead of the top-level [url](https://github.com/Homebrew/homebrew-core/blob/master/Formula/goffice.rb#L4) from gnome.org. The proposed livecheck overrides the url, but uses the generalized regex.

PR #116, opened 2019-06-21
Add livechecks

equal-l2:
Added livechecks for libcerf, rtmpdump, lzo, libvterm, libomp, and libev.


equal-l2:
All I've done is "add livechecks". How can I be better?

1480c1:
I guess edit with something like
```
Add livechecks for libcerf, rtmpdump, lzo, libvterm, libomp, and libev
```
I guess, best to just add a description with what your commits did just in case someone doesn't want to look at the commits lists for some reason/they are not expecting someone to actually have a good commit history

MikeMcQuaid:
Thanks for this and all the other PRs, @equal-l2, you rock!

PR #121, opened 2019-08-16
per-tap livecheckables

maxim-belkin:
Minimal changes to enable per-tap livecheckables.

Need advice of Ruby experts on how to better inject code into `load_formula` of `formulary.rb`.

### DIFF
--- a/livecheck/extend/formulary.rb
+++ b/livecheck/extend/formulary.rb
@@ -7,7 +7,7 @@ module Formulary
      def livecheckable_path(path)
        if (opt = path.realpath / "../../Livecheckables" / path.basename).exist?
          opt
-       elsif (opt = Pathname.new(__FILE__).realpath.dirname / "../../Livecheckables" / path.basename).exist?
+       elsif (opt = Pathname(__dir__) / "../../Livecheckables" / path.basename).exist?
          opt
        end
      end
--- a/livecheck/extend/formulary.rb
+++ b/livecheck/extend/formulary.rb
@@ -8,7 +8,7 @@ module Formulary
        path_suffix = "../../Livecheckables/#{path.basename}"
        if (opt = path.realpath/path_suffix).exist?
          opt
-       elsif (opt = Pathname(__dir__) / "../../Livecheckables" / path.basename).exist?
+       elsif (opt = Pathname(__dir__)/path_suffix).exist?
          opt
        end
      end
--- a/livecheck/extend/formulary.rb
+++ b/livecheck/extend/formulary.rb
@@ -5,10 +5,10 @@ module Formulary
     # extended to load the Livecheckable version of a formula
 
      def livecheckable_path(path)
-       if (opt1 = Pathname.new(path).realpath.dirname.dirname / "Livecheckables" / path.basename).exist?
-         opt1
-       elsif (opt2 = Pathname.new(__FILE__).realpath.dirname / "../../Livecheckables" / path.basename ).exist?
-         opt2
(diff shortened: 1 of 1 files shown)


maxim-belkin:
@youtux, are there any concerns / problems related to this PR? Per-tap livecheckables seems like a very useful thing to have...

youtux:
I agree, but I would like to have here as well somebody more experienced than me to review these changes.
@Homebrew/maintainers can you help reviewing this?

livecheck/extend/formulary.rb
```diff
@@ -3,15 +3,24 @@
 module Formulary
   class << self
     # extended to load the Livecheckable version of a formula
+
+     def livecheckable_path(path)
+       if (opt = path.realpath / "../../Livecheckables" / path.basename).exist?
```
zachauten: How is path.realpath different from Pathname(__dir__)?

maxim-belkin: `path.realpath` -- location of the formula file.
`Pathname(__dir__)` -- location of this file (`formulary.rb`)

The first `..` on line 8 "strips" the filename. The second `..` goes one directory up.

maxim-belkin: Basically, `livecheckable_path` checks whether tap has its own "Livecheckables" directory and returns it if it does exist. If not, it checks whether "Livecheckables" directory in this tap "homebrew/livecheck" exists and returns it if it does. The second check is... questionable but I think it might work well for the transition period: once per-tap livecheckables become popular, we'll be able to get rid of the "Livecheckables" directory in this tap.

MikeMcQuaid: I will do a "proper" review in the next few days but this line definitely warrants a refactor with some intermediate variables and probably comments.

MikeMcQuaid:
```suggestion
       path_suffix = "../../Livecheckables/#{path.basename}"
       if (opt = path.realpath/path_suffix).exist?
```

livecheck/extend/formulary.rb
```diff
@@ -3,15 +3,24 @@
...
+         opt
+       elsif (opt = Pathname(__dir__) / "../../Livecheckables" / path.basename).exist?
```
MikeMcQuaid:
```suggestion
       elsif (opt = Pathname(__dir__)/path_suffix).exist?
```

livecheck/extend/formulary.rb
```diff
@@ -3,15 +3,24 @@
...
       mod = Module.new
       const_set(namespace, mod)
       mod.module_eval(contents, path)
 
-      livecheckable_path = LIVECHECKABLES_PATH / path.basename
-      if File.exist?(livecheckable_path)
-        puts "Loading #{LIVECHECKABLES_PATH / path.basename}" if ARGV.debug?
-        mod.module_eval(livecheckable_path.read, livecheckable_path)
+      lc_path = livecheckable_path(path)
+      if File.exist?(lc_path)
```
MikeMcQuaid:
```suggestion
      if lc_path.exist?
```

MikeMcQuaid:
> I am wondering about the overall approach of Livecheckables to "override" (replicate and modify) the `load_formula` method -- it does not feel ~wright~ ~write~ right. What would be the proper thing to do here?

You could have a separate loader which does an `include` or `extend` onto the instance after it's been loaded by the Formula loader?

MikeMcQuaid:
Merging for now as looks good but any additional refactoring would be 🥇

PR #126, opened 2019-09-02
euristic.rb: escape package name for gnome regex (gtk+3)

jeffreywildman:
Before:

```bash
$ brew livecheck gtk+3
Error: Unable to get versions for gtk+3
```

After:

```bash
$ brew livecheck gtk+3
gtk+3 (guessed) : 3.24.10 ==> 3.94.0
```


MikeMcQuaid:
Thanks again @jeffreywildman!

PR #128, opened 2019-09-17
swi-prolog: move to https url

Moisan:
The current url creates a redirection and leads to the following error: 
```
Error: redirection forbidden: http://www.swi-prolog.org/download/stable/src -> https://www.swi-prolog.org/download/stable/src
```


MikeMcQuaid:
Thanks again @Moisan!

PR #129, opened 2019-09-21
systemc: use https url

Moisan:
The previous url redirected to the https one. This led to a `redirection forbidden` error.


MikeMcQuaid:
Thanks again @Moisan! I've given you commit access to this repository.

PR #131, opened 2019-10-12
Support upstreams that tag only with 'debian/' prefix

dawidd6:
There are upstreams that only do `debian/` prefixed tags in their repositories, like this one: https://github.com/agx/git-buildpackage.

This PR aims to support checking the life of formulae that pull from repositories like these.

### DIFF
--- a/livecheck/euristic.rb
+++ b/livecheck/euristic.rb
@@ -35,9 +35,22 @@ def version_euristic(urls, regex = nil)
     when DownloadStrategyDetector.detect(url) <= GitDownloadStrategy
       puts "Possible git repo detected at #{url}" if ARGV.debug?
 
-      git_tags(url, regex).each do |tag|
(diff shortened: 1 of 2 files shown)


livecheck/euristic.rb
```diff
@@ -35,9 +35,22 @@ def version_euristic(urls, regex = nil)
...
+      tags = git_tags(url, regex)
+      tags_only_debian = true
+
+      # Check if upstream only does 'debian/' prefixed tags
+      tags.each do |tag|
+        if tag !~ /debian\//
+          tags_only_debian = false
+          break
+        end
+      end
```
Moisan: Could you move this to a separate method? Testing is almost non-existent in livecheck but that would make things easier for future tests while also trying to reduce the length of this already huge function.

dawidd6: It is done!

PR #134, opened 2019-10-28
euristic.rb: replace if.nil? with ||= assignments

maxim-belkin:
This PR replaces this:

```ruby
if regex.nil?
  regex = ...
end
```

with this:

```ruby
regex ||= ...
```

Nothing more...

### DIFF
--- a/livecheck/euristic.rb
+++ b/livecheck/euristic.rb
@@ -64,9 +64,7 @@ def version_euristic(urls, regex = nil)
              "at #{url}"
       end
 
-      if regex.nil?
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Ping @youtux.
Could you please have a look and tag others as necessary.

Moisan:
LGTM but that's an example of when tests would be useful to double check we didn't break anything.

Moisan:
And thank you for your contribution! homebrew-livecheck codebase needs love and little change like this make it better.

PR #135, opened 2019-11-03
Provide new --tap option

dawidd6:
With `--tap` option, one can check all formulae from specified tap.

For example: `brew livecheck --tap=linuxbrew/xorg` would check all
formulae available in this tap.


Moisan:
LGTM. Good feature idea too!

PR #137, opened 2019-11-28
exiftool: update url

samford:
The exiftool livecheck is failing with `Error: redirection forbidden: https://www.sno.phy.queensu.ca/~phil/exiftool/history.html -> http://exiftool.org/history.html`, so this addresses the issue.


Moisan:
Thank you again @samford !

PR #138, opened 2019-11-28
icu4c: update url and regex

samford:
https://ssl.icu-project.org URLs appear to be dead or in a state of prolonged downtime.  http://ssl.icu-project.org (not using https) simply redirects to http://site.icu-project.org

With this in mind, it may be a good idea to modify the icu4c livecheckable to use a different page.  I'm using the [ICU download page](http://site.icu-project.org/download/) in this PR but that site doesn't allow for secure connections, unfortunately.

If a secure connection is preferred, it's also possible to use the GitHub release page like so:

```ruby
  livecheck :url => "https://github.com/unicode-org/icu/releases",
            :regex => /href="\/unicode-org\/icu\/releases\/tag\/release-[^"]+"[^>]*>ICU ([\d\.]+)</
```

The main downside of using the GitHub release page is that this regex relies on the release title always being "ICU" and then the version (e.g., "ICU 65.1").  This is true for newer releases but older releases simply use the name from the tag, like "release-60-2".  Using the release title is necessary because the periods in the version string are replaced with hyphens or underscores elsewhere on the page (tags, filenames).  It's also worth noting that this GitHub regex targets non-RC releases (e.g., "ICU 65 RC" won't be matched), as I assumed that would be desired behavior.

What are everyone's thoughts on this?

### DIFF
--- a/Livecheckables/icu4c.rb
+++ b/Livecheckables/icu4c.rb
@@ -1,4 +1,4 @@
 class Icu4c
-  livecheck :url => "https://ssl.icu-project.org/files/icu4c/",
-            :regex => /href="([\d.]+\.[\d.]+\.?[\d.]*)\/"/
+  livecheck :url => "https://github.com/unicode-org/icu/releases",
+            :regex => /href="\/unicode-org\/icu\/releases\/tag\/release-[^"]+"[^>]*>ICU ([\d\.]+)</
 end


Moisan:
> The main downside of using the GitHub release page is that this regex relies on the release title always being "ICU" and then the version (e.g., "ICU 65.1"). This is true for newer releases but older releases simply use the name from the tag, like "release-60-2"

Since the goal of livecheck is to check for the latest release, I don't think we need to worry about the naming of older releases. Also there has been some effort to make sure we use only https urls so I would prefer the Github one.

samford:
The branch now contains the GitHub version, so it should be good to go.

While I have your ear, I've been working on a number of livechecks and was wondering if there's any preference on how to submit them.  Normally I would create separate PRs for each formula (e.g., like homebrew-core) but if you would prefer for me to bundle them into one PR (to make it easier than having to merge each PR individually), then I could do that.

Moisan:
Thank you for this PR! 

> Normally I would create separate PRs for each formula (e.g., like homebrew-core) but if you would prefer for me to bundle them into one PR (to make it easier than having to merge each PR individually), then I could do that.

One PR per formula is easier to review and merge so I would stick with that.

PR #142, opened 2019-12-01
libsamplerate: add livecheck

samford:
Unfortunately, a secure connection isn't available and there doesn't appear to be an alternative location to get release information from (e.g., the GitHub repo doesn't have releases).


Moisan:
Thank you @samford ! https links are preferred but when the Homebrew formula doesn't (yet) use one, it's fine for the livecheckable to do the same.

PR #143, opened 2019-12-01
libsndfile: add livecheck

samford:
The main website (used in this commit) doesn't provide a secure connection, unfortunately.   Release versions *are* available in the [GitHub repo](https://github.com/erikd/libsndfile/releases) but the last release was over two years ago, so it's hard to say whether the GitHub releases page will be updated in the future.  However, if you prefer to use the GitHub releases page, I can update the livecheck accordingly.


Moisan:
Thanks @samford ! The goal of livecheck is to find the most up to date release. In this case, going with the same website as the current Formula makes sense.

PR #146, opened 2019-12-01
opencore-amr: add livecheck

samford:
opencore-amr doesn't seem to work properly with the default Sourceforge version-finding (even after updating the URL in the formula), so I've created a livecheck to handle this special case.


Moisan:
I'm worried that without this PR `brew livecheck opencore-amr` returns this error:
```
Error: undefined method `[]' for nil:NilClass
```
A bit more specific message would be nice, but that's something for another PR. Thank you @samford !

samford:
I certainly agree about that error.  Without knowing the formula that causes the error, you have to hunt it down based on the formulae that have run before/after the error, which isn't so great.

For what it's worth, there were 47 other formulae that encountered that error when I did a full run a week ago.

PR #147, opened 2019-12-01
Update SourceForge RSS URL

samford:
The URLs for SourceForge project RSS feeds changed (older URLs were redirecting to the new URLs), so this brings the URL format up to date.


Moisan:
Good catch, thank you @samford !

PR #161, opened 2019-12-15
euristic.rb: support checking pypi packages

dawidd6:
Example formula that benefits from this change: `internetarchive`.


Moisan:
LGTM. After merging I'll do a quick round in the Livecheckables, this probably replace many of them. Thank you @dawidd6 !

PR #177, opened 2019-12-31
Github action to run brew livecheck on added/modified Livecheckables during PRs

Moisan:
Whenever a PR adds or modifies a Livecheckable, I manually check that it works locally. With this PR, the CI will do that check.

@MikeMcQuaid : requesting your review since I am not that familiar with Github actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,4 +26,4 @@ jobs:
       run: brew livecheck
 
     - name: Run brew livecheck on added/modified Livecheckables
-      run: git diff --name-only --diff-filter=AM master -- Livecheckables/ | sed "s/Livecheckables\/\(.*\).rb/\1/" | xargs brew livecheck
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -24,3 +24,6 @@ jobs:
 
     - name: Run brew livecheck
       run: brew livecheck
+
+    - name: Run brew livecheck on added/modified Livecheckables
+      run: git diff --name-only --diff-filter=AM master -- Livecheckables/ | sed "s/Livecheckables\/\(.*\).rb/\1/" | xargs brew livecheck
```
MikeMcQuaid:
```suggestion
      run: git diff --name-only --diff-filter=AM origin/master -- Livecheckables/ | sed "s|Livecheckables/\(.+\)\.rb|\1|" | xargs brew livecheck
```

PR #181, opened 2020-01-02
ARGV: replace usages of ARGV.formulae with Homebrew.args

Moisan:
In the spirit of the [ARGV cleanup](https://github.com/Homebrew/brew/issues/5730), replace usage of `ARGV.formulae` with `Homebrew.args.formulae`.


MikeMcQuaid:
Are you sure this works? I think you'll need to run `Homebrew::CLI::Parser.new do` for this to work.

PR #193, opened 2020-01-31
brew-livecheck: allow checking installed HEAD only formulae

dawidd6:
With this change, it becomes possible to livecheck the HEAD only
formulae if they are installed. If they are not - inform the user about
it.

### DIFF
--- a/cmd/brew-livecheck.rb
+++ b/cmd/brew-livecheck.rb
@@ -44,8 +44,13 @@ def print_latest_version(formula)
     return
   end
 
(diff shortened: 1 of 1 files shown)


Moisan:
I couldn’t find an example an a formula that is head-only to test this. Could you point me to one?

dawidd6:
I used a formula from my tap to test it: https://github.com/dawidd6/homebrew-tap/blob/master/Formula/forgit.rb

PR #198, opened 2020-02-07
livecheck: fix early return

dawidd6:
`brew livecheck FORMULA` always exits doing nothing, without this change.


Moisan:
Yep that fixes the issue. I'll merge this without @MikeMcQuaid review since livecheck is currently broken without it.

MikeMcQuaid:
Thanks for the quick fix!

PR #199, opened 2020-02-07
livecheck: add possibility to output in JSON

dawidd6:
This PR adds a new `--json` option.
If passed, `livecheck` will output the informations in JSON format.


Moisan:
Very good idea! The only piece of information that is not included in the JSON is `is_newer_than_upstram` (yes the code includes a typo). Could you add it?

dawidd6:
Done. I also fixed a typo (`upstram` -> `upstream`).

PR #206, opened 2020-02-17
azure-cli: add livecheckable

samford:
This adds a livecheckable for `azure-cli` since the formula uses an unusual URL and doesn't livecheck properly.


Moisan:
Thank you @samford !

PR #209, opened 2020-02-17
luajit: add livecheckable

samford:
The default livechecking for LuaJIT doesn't distinguish between stable and beta releases (2.1.0-beta3 is given as the latest version), so this adds a livecheckable to address this.

This assumes that `livecheck` should return the latest stable version and that alpha/beta/RC/etc. releases aren't desirable, of course.


Moisan:
> This assumes that livecheck should return the latest stable version and that alpha/beta/RC/etc. releases aren't desirable, of course.

Yes this is currently the intent of livecheck. 

Thank you @samford !

PR #212, opened 2020-02-17
tcl-tk: add livecheckable

samford:
The default `tcl-tk` livecheck lists an alpha version as the latest (9.0a1), so this creates a livecheckable to identify the latest stable version.


Moisan:
Thank you @samford !

PR #217, opened 2020-02-18
composer: add regex

samford:
composer's existing livecheckable doesn't filter Git tags, so it also picks up release candidate versions (e.g., 1.10.0-RC). This adds a regex that restricts livecheck to stable releases.


Moisan:
Thank you @samford !

PR #218, opened 2020-02-18
Restrict gnome.org matching

samford:
The formulae to which the gnome.org behavior in `livecheck/heuristic.rb` should apply all have a download.gnome.org URL, so I've modified the URL in `heuristic.rb` accordingly.

The default `easy-git` livecheck fails (``Error: easy-git: undefined method `[]' for nil:NilClass``) because its URL uses the people.gnome.org domain and the gnome.org behavior is wrongly applied. Further, the gnome.org behavior even overrides the livecheckable that I created to address this issue, so `heuristic.rb` needed to be updated for accuracy but also to allow for `easy-git` to be fixed with a livecheckable.

### DIFF
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -123,7 +123,7 @@ def version_heuristic(urls, regex = nil)
         version = Version.new(match)
         match_version_map[match] = version
       end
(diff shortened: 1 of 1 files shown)


Moisan:
LGTM! Thanks!

PR #235, opened 2020-02-23
anime-downloader: add livecheckable

nandahkrishna:
Adding Livecheckable for anime-downloader.

Do review and let me know if I should make any changes!

### DIFF
--- a/Livecheckables/anime-downloader.rb
+++ b/Livecheckables/anime-downloader.rb
@@ -1,0 +1,4 @@
+class AnimeDownloader
+  livecheck :url => "https://www.github.com/vn-ki/anime-downloader/releases",
+            :regex => %r{href="/vn-ki/anime-downloader/tree/([0-9\.])+"}
+end


dawidd6:
```
Error: anime-downloader: Unable to get versions
```

Doesn't seem to work.

nandahkrishna:
Oh alright. Thought it'd be useful to have all formulae here, as I intend to take up the `livecheck` DSL issue. Let me know if I have to close this!

dawidd6:
If a project is hosted on any known site (like Github, Gitlab or GNU) there is no need to create a dedicated livecheckable for it. Unless `brew livecheck` picks up `-rc` or other unwanted releases, which is not the case in this situation.

nandahkrishna:
Alright, I'll keep that in mind. Thanks.

samford:
@nandahkrishna If you would like to better understand how `livecheck` works, take a look at the code in the livecheck folder in this repo, especially [heuristic.rb](https://github.com/Homebrew/homebrew-livecheck/blob/master/livecheck/heuristic.rb).

`livecheck` handles a number of formulae automatically and we only add manual livechecks when the default behavior is inadequate.

nandahkrishna:
Ah I see. I now understand what @dawidd6 was pointing out about the popular sites.

As I mentioned in #236 there are errors on using `livecheck` for some formulae, and I see that these are hosted on sites other than those included in heuristic.rb; I'll make PRs to try and fix those alone for now. 

Thanks @samford, that was insightful.

PR #237, opened 2020-02-23
[WIP] Adding livecheckables for formulae

nandahkrishna:
Refer to Issue #236, adding explicit Livecheckables for formulae where `brew livecheck <formula>` throws errors (or) shows wrong version names.

PR #238, opened 2020-02-23
aamath: add livecheckable

nandahkrishna:
Adding live checkable for `aamath` as per #237.

### DIFF
--- a/Livecheckables/aamath.rb
+++ b/Livecheckables/aamath.rb
@@ -1,0 +1,3 @@
+class Aamath
+  livecheck :regex => /aamath-(\d+(?:\.\d+)+)\.tar/
+end


nandahkrishna:
Modified description.

samford:
I've looked through your submitted PRs and I have some tips and feedback. I'll leave the more general feedback in this PR along with specific items to address in each PR. Your livecheckables all worked when I tested them but there are some areas that can potentially improve and I wanted to try to guide you in the right direction at the outset.

## `:url` and `:regex` are not always required together
One thing to note about livecheckables is that the URL isn't always necessary and in other cases the regex isn't necessary. Due to `livecheck`'s heuristic, you can sometimes get away with only providing a regex or only providing a URL. I only speak for myself but I try to keep livecheckables as minimal as possible and lean on the heuristic (as this may be easier to maintain in the future). With the `aamath` livecheckable, you can omit the URL, as `livecheck` is already using the one you provided.

## Use the `--debug` (`-d`) flag
When you're working on a livecheckable, it's good to run the livecheck for the formula with the debug flag, to see what `livecheck` is doing behind the scenes: `brew livecheck --debug aamath`. The output will tell you where `livecheck` is looking for versions and you may only need to provide a regex or a URL.

## Make sure the regex syntax matches your intention
Be sure you escape periods in the regex where they are meant to be matching periods. Unescaped periods mean "match anything".

## Generalize the regex when possible
When you're matching a tar.gz/tar.bz2/tar.xz, etc. file, write the regex so it's simply matching ".tar" and leave off the final part of the extension (.gz, .bz2, .xz, etc.).  This will allow the regex to continue matching the files without needing further modification even if the extension changes. When writing a livecheckable regex, I often ask myself, "What could change upstream that would break this regex?" or "How can I make this more general or less fragile without introducing false positives/negatives?"

## Try to look for prevailing standards or areas for improvement
Lastly, I've started using a more specific regex for matching versions like 1.2, 1.2.3, v1.2.3, etc. rather than the more generic `[0-9\.]+` that would technically match any number of bizarre aberrations like ...., 12345, ...2....3., etc. The regex I've been using is a modification of one found in the livecheck code: `(\d+(?:\.\d+)+)` You can mix this into the regex like: `/aamath-(\d+(?:\.\d+)+)\.tar/`

## Closing
So, if we combine all of this advice together, the livecheckable ends up looking like:

```ruby
class Aamath
  livecheck :regex => /aamath-(\d+(?:\.\d+)+)\.tar/
end
```

I hope this is helpful information. I think I need to try to carve out some time to write up documentation for potential contributors to `livecheck` but I'll try to provide feedback like this as people submit PRs. Let me know if you have any additional questions.

nandahkrishna:
Thank you so much @samford for the guide. I could probably put this into a contributing guide through another PR if you'd like me to. I'll make the changes you've suggested here and in the other PRs and get back to you. Once again, thanks a ton for this!

SMillerDev:
Adding this to documentation would be good, nice writeup @samford

nandahkrishna:
Modified `aamath.rb` as per the suggestions.

samford:
> I could probably put this into a contributing guide through another PR if you'd like me to.

I appreciate the offer but I'm planning on reworking this comment into docs as part of a more comprehensive approach to improving the experience for new contributors to homebrew-livecheck.

Thanks for incorporating the feedback into your PR and thank you as well for your contributions to `livecheck`!

PR #239, opened 2020-02-23
a52dec: add livecheckable

nandahkrishna:
Adding livecheckable for `a52dec`.


samford:
The only feedback for this PR is that you may want to use the previously-mentioned regex for version matching. It would be `/a52dec-(\d+(?:\.\d+)+)/` in this case.

nandahkrishna:
Done.

PR #240, opened 2020-02-23
aacgain: add livecheckable

nandahkrishna:
Adding livecheckable for `aacgain`.

### DIFF
--- a/Livecheckables/aacgain.rb
+++ b/Livecheckables/aacgain.rb
@@ -1,0 +1,4 @@
+class Aacgain
+  livecheck :url => "https://aacgain.altosdesign.com/alvarez/",
+            :regex => /aacgain-(\d+(?:\.\d+)+)\.tar/
+end


samford:
The URL in this livecheckable redirects to the same URL with a trailing slash. It's good to test the URL before adding it to the livecheckable, to avoid unnecessary redirections. You can do this by navigating to the page in your browser or you can use `curl -I -X GET https://www.example.com` to view the response headers for a URL.  In this case, it's a 301 (redirection) and provides the target URL in the "location" header.

* The periods in the extension need to be escaped.
* Match ".tar", rather than ".tar.bz2", to be more general.
* Consider using the common version matching regex. In this case, it would be: `/aacgain-(\d+(?:\.\d+)+)\.tar/`

nandahkrishna:
Oh okay. Thanks for the suggestion! Also, the livecheckable has been updated.

PR #241, opened 2020-02-23
abuse: add livecheckable

nandahkrishna:
Adding livecheckable for `abuse`.


samford:
* The periods in the extension need to be escaped
* Match ".tar", rather than ".tar.bz2", to be more general
* Consider using the common version matching regex. In this case, it would be: `/abuse-(\d+(?:\.\d+)+)\.tar/`

nandahkrishna:
Done.

PR #242, opened 2020-02-23
aardvark_shell_utils: add livecheckable

nandahkrishna:
Adding livecheckable for `aardvark_shell_utils` to fix `brew livecheck` errors.


samford:
* The periods in the extension need to be escaped
* Match ".tar", rather than ".tar.gz", to be more general
* Consider using the common version matching regex. In this case, it would be: /aardvark_shell_utils-(\d+(?:\.\d+)+)\.tar/

nandahkrishna:
Done.

PR #243, opened 2020-02-23
lrzip: add livecheckable

nandahkrishna:
Adding livecheckable for `lrzip` to fix `brew livecheck` errors.


samford:
The only feedback I have for this is that it may (or may not) be better to match version strings in the filenames rather than using the "Latest-Is-..." file. If that file is ever deleted or if they forget to update it, this would fail to detect the new version.

Like the other PRs, we want to more specific version matching regex. With these two items in mind, the regex would be: `/lrzip-(\d+(?:\.\d+)+)\.tar/`

nandahkrishna:
Done.

PR #245, opened 2020-02-24
Update github action with current branch name during diff

Moisan:
I've spotted a few PR where the CI would fail if a newer change was made in master but not yet rebased on the PR branch. This forces `git diff` to only check newer changes coming from the PR branch.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,12 +27,9 @@ jobs:
 
     - name: Fetch origin/master
       run: git fetch --depth=1 origin master
-    
-    - name: Get branch name
-      uses: nelonoel/branch-name@v1
 
     - name: Run brew livecheck on added/modified Livecheckables
       run: |
-          git diff --name-only --diff-filter=AM origin/master...${BRANCH_NAME} -- Livecheckables/ |
+          git diff --name-only --diff-filter=AM origin/master...${{ github.head_ref }} -- Livecheckables/ |
           sed "s|Livecheckables/\(.*\)\.rb|\1|" |
           xargs brew livecheck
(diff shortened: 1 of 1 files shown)


Moisan:
`${GITHUB_REF##*/}` does not seem to be the branch name in Github action. I'm looking for another way to obtain that information.

nandahkrishna:
I believe `github.event.pull_request.head.ref` might be what you're looking for. It's part of the `github` context in Actions. I also think `github.head_ref` may work, since the triggering event is a pull request.

Moisan:
Nope, still not okay:
```
Run git diff --name-only --diff-filter=AM origin/master...${github.head_ref} -- Livecheckables/ |
  git diff --name-only --diff-filter=AM origin/master...${github.head_ref} -- Livecheckables/ |
  sed "s|Livecheckables/\(.*\)\.rb|\1|" |
  xargs brew livecheck
  shell: /bin/bash -e {0}
/Users/runner/runners/2.165.2/work/_temp/694fec95-047f-40ad-940a-31a03bec9119.sh: line 1: origin/master...${github.head_ref}: bad substitution
```

nandahkrishna:
My mistake, I should've mentioned that it would have to be `${{ github.head_ref }}`, at least according to the [docs](https://help.github.com/en/actions/reference/contexts-and-expression-syntax-for-github-actions#github-context). Hope your current fix works though.

Moisan:
Let's try that then, I prefer to use the default Github reference to an external action.

nandahkrishna:
This seems to fail as well, substituting `github_action_branch_name`. Sorry about this, I'll check the docs again and see if there's an alternative from GitHub. This might even be setting the value of github.head_ref to an environment variable.

nandahkrishna:
Might using the commit SHA via `github.sha` (commit SHA that triggered the workflow run) work in this case? Unfortunately I've got no way to test this locally except just executing a `git diff` and trying to mimic the requirement here.

Moisan:
That did the trick, thanks @nandahkrishna !

nandahkrishna:
Glad it worked!

PR #247, opened 2020-02-24
aalib: add livecheckable

nandahkrishna:
This is not yet complete, as I have a query. Formula `aalib` is currently at version `1.4rc5`, which is an `rc` version. According to #246, if I understand correctly, such versions are not allowed on Homebrew/core, but `aalib` seems to be an exception (possibly more exist). How must I handle this?

Output for `brew livecheck aalib`:
* Before this change
```Error: aalib: undefined method `[]' for nil:NilClass```
* After this change
`aalib : 1.4rc5 ==> 1.2`

### DIFF
--- a/Livecheckables/aalib.rb
+++ b/Livecheckables/aalib.rb
@@ -1,0 +1,4 @@
+class Aalib
+  livecheck :url => "https://sourceforge.net/projects/aa-project/files/aa-lib/",
+            :regex => /aalib-(\d+(?:\.\d+)+.*?)\.tar/
+end


dawidd6:
`aalib` seems to be a very old project. Latest release was made in 2001 and it's a `rc` one.

And it was added to Homebrew a long time ago too.

nandahkrishna:
Alright, I've modified the regex accordingly.

samford:
If you're ever curious about the history of a formula, you can hop into the `homebrew-core` repo (`cd $(brew --repo homebrew/core)`) and use [`git log -L` to view previous commits affecting specific lines](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt--Lltstartgtltendgtltfilegt), such as the `url` line. As Dawid hinted at, this formula was originally added at version 1.4rc4 in 2009, so it's an exception to the normal rule of targeting stable releases.

One small bit of feedback: outside of the main capture group, you want to make the groups within it non-capturing by using `?:` after the opening parenthesis. For example, the `(rc\d+)?` part of the regex would be `(?:rc\d+)?`. This helps to keep things a little cleaner internally.

Do we only want to restrict this to stable and RC releases or would we also use potential alpha, beta, etc. releases for `aalib`? In the latter case, we can just allow the regex to accept anything after the version, like: `/aalib-(\d+(?:\.\d+)+.*?)\.tar/` What do you think, @dawidd6?

dawidd6:
I don't think this project will get any new releases, but I would say the less strict regex is better.

nandahkrishna:
I'll change it to the less strict regex then.

PR #286, opened 2020-03-01
Fix RuboCop Layout/HashAlignment offenses

samford:
This aligns hash keys and values throughout livecheck (though primarily in livecheckables), in line with RuboCop's [Layout/HashAlignment](https://rubocop.readthedocs.io/en/latest/cops_layout/#layouthashalignment) cop.


Moisan:
The CI is testing every modified Livecheckable so unless we ensure they all work before fixing the linting, the CI will always fail.

samford:
> The CI is testing every modified Livecheckable so unless we ensure they all work before fixing the linting, the CI will always fail.

I was in the process of writing a comment to this very effect, ahah. I cancelled the workflow a moment before your comment popped up.

This will be fine to merge anyway, since the changes are purely stylistic rather than functional.

PR #296, opened 2020-03-02
deployer: update regex

samford:
The existing livecheckable for `deployer` gives an error (`Unable to get versions`), so this updates the regex to fix the issue.


samford:
For anyone stumbling across this in the future, apparently I created two PRs for the same change at different times. The second commit looks blank in master because nothing was effectively changed after the first commit. Between the two commits, I troubleshoooted a CI bug and wrote up a lengthy issue, so I had forgotten about the previous merge (as I hadn't updated my local master branch yet).

This kind of thing sometimes happens when you put too many hours into a project in one day 😞

PR #303, opened 2020-03-04
Allow :skip in livecheckables to skip formulae

samford:
Some formulae are important and we want to keep them around but they will not see a new version anytime soon (if ever) and livecheck shouldn't waste effort checking them. For example, the `imap-uw` formula states, "imap-uw is unmaintained software; the author has passed away and there is no active successor project."

I personally have this installed and I feel that there's still value in keeping this in Homebrew but the URLs in the formula are dead (imap-uw has been seemingly scrubbed from University of Washington's website). As it stands, livecheck will always fail with this formula unless something changes upstream. I regularly run `brew livecheck --installed` and `brew livecheck --all`, so it would be nice to avoid unnecessary checks.

To address this issue, this PR adds support for using `:skip` in the `livecheck` arguments to tell livecheck not to check for new versions of a formula. The value is either a string containing a _very_ brief message explaining why the formula is skipped or `nil` (no message).

The livecheckable looks like this in practice:

```ruby
class ImapUw
  livecheck :skip => "Not maintained"
end
```

The output looks like this when an explanation is provided:

```
imap-uw : skipped - Not maintained
```

The output looks like this when an explanation is omitted (i.e., `:skip => nil`):

```
imap-uw : skipped
```

Please look through the code and provide feedback if you have a moment. My Ruby's a bit rusty these days and my knowledge of Homebrew is incomplete, so there's likely room for improvement.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -99,6 +99,21 @@ module Homebrew
       return
     end
 
+    has_hash_arg_with_skip =
+      formula.livecheck_args.is_a?(Hash) && formula.livecheck_args.key?(:skip)
+    if has_hash_arg_with_skip || formula.livecheck_args == :skip
+      if has_hash_arg_with_skip &&
+         formula.livecheck_args[:skip].is_a?(String) &&
+         !formula.livecheck_args[:skip].empty?
+        skip_msg = " - #{formula.livecheck_args[:skip]}"
+      else
+        skip_msg = ""
+      end
+
+      puts "#{Tty.red}#{formula}#{Tty.reset} : skipped#{skip_msg}" unless Homebrew.args.quiet?
+      return
+    end
+
     current = formula.stable? ? formula.version : formula.installed_version
     latest = formula.stable? ? formula.latest : formula.latest_head_version
     if (m = latest.to_s.match(/(.*)-release$/)) && !current.to_s.match(/.*-release$/)
--- a/livecheck/extend/formula.rb
+++ b/livecheck/extend/formula.rb
@@ -5,12 +5,16 @@ class Formula
     self.class.latest
   end
 
+  def livecheck_args
+    self.class.livecheck_args
+  end
+
   def livecheckable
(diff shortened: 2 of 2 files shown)


dawidd6:
Good stuff. Can we also somehow make it to return JSON if `--json` is passed?

samford:
> Good stuff. Can we also somehow make it to return JSON if `--json` is passed?

That's a systemic issue with livecheck and I aim to fix it in a future PR (addressing `Error:`, `warning:`, `fatal:`, `versioned`, `skipped`, etc). It's something that requires discussion about how to handle the errors with respect to the JSON output format, so I wanted to keep this PR simple and defer it until later when I work on the other issues.

cmd/livecheck.rb
```diff
@@ -99,6 +99,19 @@ def print_latest_version(formula)
       return
     end
 
+    if formula.livecheck_args.class == Hash &&
```
Moisan: I think the `formula.livecheck_args.is_a?(Hash)` is clearer. The parentheses is to make sure the rest of the if statement is not included in the `is_a?`.

cmd/livecheck.rb
```diff
@@ -99,6 +99,19 @@ def print_latest_version(formula)
...
+       formula.livecheck_args.key?(:skip)
+      if formula.livecheck_args[:skip].class == String &&
```
Moisan: Same here with `is_a?`.

cmd/livecheck.rb
samford: Ah, yes, _that_'s the Ruby way. Thanks for the heads up.

samford:
> We could also allow the use of something like `livecheck :skip` by checking if `livecheck_args == :skip` and setting `skip_msg = ""` in those cases.

Thanks for the reminder.  I meant to support that case but apparently forgot.  I don't want this to encourage skipping livecheckables without providing a brief message but I _do_ like to make things flexible when I can.

Issue #300, opened 2020-03-03
GNOME releases with an odd minor version are development releases


Bo98:
```
$ brew livecheck atk
atk (guessed) : 2.34.1 ==> 2.35.1
```

This goes against the audit check, which disallows 2.35.x: https://github.com/Homebrew/brew/blob/bcb8f4da5f48bb4ad96c1c66af0a3b5996f711d1/Library/Homebrew/dev-cmd/audit.rb#L782

Note however that there are exceptions: https://github.com/Homebrew/brew/blob/bcb8f4da5f48bb4ad96c1c66af0a3b5996f711d1/Library/Homebrew/dev-cmd/audit.rb#L751-L756

samford:
I've been meaning to get around to fixing this. I had noticed this issue in the past and created a tentative solution but I haven't yet incorporated it into livecheck because I was looking for all the formulae that shouldn't be restricted in this manner. I had found a few myself but, last I left this, I hadn't gotten around to finding them all.

It looks like the `gnome_devel_whitelist` is exactly what I was looking for. Is it possible to access that array from within livecheck's heuristic or would I have to manually duplicate it?

I'll work on getting this together into a PR when I get a chance in a bit.

Bo98:
> Is it possible to access that array from within livecheck's heuristic or would I have to manually duplicate it?

As it currently is, you can't access it directly. However having to update two repos constantly for exceptions would suck. I'm sure something can be worked out.

With that said, it's worth noting that entries on that list are currently version pinned. I don't know if there's a particular reason for that. For example, if `gtk-mac-integration` 2.1.4 came out tomorrow, should it be updated?

samford:
It sounds like we'll just have to live with having this information in two places for the moment and address it in the future if/when we can. I've run out of time today but I'll try to work on integrating this with my existing solution and getting a PR up in the next day or two.

I often see @tschoonj working on GNOME-related formulae, so he may have an idea about why those formulae are pinned to specific versions.

PR #305, opened 2020-03-05
heuristic: add gnome.org version restrictions

samford:
As discussed in #300, the GNOME stategy  doesn't restrict matching to stable versions (i.e., using the "even-numbered minor versions are stable" scheme).  Some GNOME versions also use x.90+ as development versions, so this wrinkle is also accounted for in the regex.

This PR restricts the versions in the GNOME strategy by default, unless the package name is in the `gnome_devel_whitelist` array. The array members here are borrowed from `Homebrew/dev-cmd/audit.rb`, omitting the version numbers and `pygtkglext` (which has been deleted from Homebrew), and renaming `libart` to correspond with the package name (`libart_lgpl`).

This fixes version-matching for nearly all GNOME-related formulae but some of these will need subsequent livecheckables to address additional issues:

* gtk+: Restrict to 2.x versions
* gtk+3: Restrict to 3.x versions
* gtkmm: Restrict to 2.x versions
* gtkmm3: Restrict to 3.x versions
* gtksourceview: Restrict to 2.x versions
* gtksourceview3: Restrict to 3.x versions
* gtksourceview4: Restrict to 4.x versions
* gtksourceviewmm: Restrict to 2.x versions
* gtksourceviewmm3: Restrict to 3.x versions
* libgit2-glib: Needs to be forced to use the GNOME strategy (via URL, until `:using` is implemented); I'll create a PR after this one is merged
* libxml++: Restrict to 2.x versions
* pygobject3: Restrict to 3.x versions

I initially created a hacky way of automatically handling the major version restrictions and, though it worked, I wasn't happy with it. I felt it would be better to add the ability to specify major/minor/patch numbers to restrict matching to and _then_ create the above livecheckables. I'll likely work on this feature next, since it will probably be relatively easy and it's useful beyond just these GNOME formulae.

Anyway, if anyone who sees this will review the code and provide feedback, that would be appreciated.

### DIFF
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -2,6 +2,15 @@ require "utils"
 
 # TODO: distinguish between a filter regex and a matching regex
 def version_heuristic(livecheckable, urls, regex = nil)
+  # Formulae that do not use GNOME's "even-numbered minor is stable" scheme
+  gnome_devel_whitelist = [
+    "gcab",
+    "gtk-doc",
+    "gtk-mac-integration",
+    "libart_lgpl", # The package name for libart is "libart_lgpl"
+    "libepoxy",
+  ].freeze
+
   urls.each do |url|
     # Skip Gists until/unless we create a method of identifying revisions
     next if url.include?("gist.github.com")
@@ -144,7 +153,12 @@ def version_heuristic(livecheckable, urls, regex = nil)
         puts "Possible GNOME package [#{package}] detected at #{url}"
       end
 
-      regex ||= /#{Regexp.escape(package)}-([\d.]+\.[\d.]+\.[\d.]+)\.t/
+      # Restrict versions to even numbered minor versions (except x.90+)
+      if gnome_devel_whitelist.include?(package)
+        regex ||= /#{Regexp.escape(package)}-(\d+(?:\.\d+)+)\.t/
+      else
+        regex ||= /#{Regexp.escape(package)}-(\d+\.([0-8]\d*?)?[02468](?:\.\d+)*?)\.t/
+      end
 
       page_matches(page_url, regex).each do |match|
         version = Version.new(match)


chenrui333:
Nice, I will check out tomorrow, too late right now for me. :)

livecheck/heuristic.rb
```diff
@@ -144,7 +153,12 @@ def version_heuristic(livecheckable, urls, regex = nil)
         puts "Possible GNOME package [#{package}] detected at #{url}"
       end
 
-      regex ||= /#{Regexp.escape(package)}-([\d.]+\.[\d.]+\.[\d.]+)\.t/
+      # Restrict versions to even numbered minor versions (except x.90+)
+      if gnome_devel_whitelist.include?(package)
+        regex ||= /#{Regexp.escape(package)}-(\d+(?:\.\d+)+)\.t/
+      else
+        regex ||= /#{Regexp.escape(package)}-(\d+\.([0-8]\d*?)?[02468](?:\.\d+)*?)\.t/
```
Moisan: I think `(\d+\.[02468](?:\.\d+)*?)\.t` is equivalent. What is the purpose of `([0-8]\d*?)?`?

livecheck/heuristic.rb
```diff
@@ -2,6 +2,15 @@
...
 # TODO: distinguish between a filter regex and a matching regex
 def version_heuristic(livecheckable, urls, regex = nil)
+  # Formulae that do not use GNOME's "even-numbered minor is stable" scheme
+  gnome_devel_whitelist = [
+    "gcab",
+    "gtk-doc",
+    "gtk-mac-integration",
+    "libart_lgpl", # The package name for libart is "libart_lgpl"
+    "libepoxy",
+  ]
```
Moisan: Could you move this near the Gnome url section?

livecheck/heuristic.rb
```diff
@@ -144,7 +153,12 @@ def version_heuristic(livecheckable, urls, regex = nil)
         puts "Possible GNOME package [#{package}] detected at #{url}"
       end
 
-      regex ||= /#{Regexp.escape(package)}-([\d.]+\.[\d.]+\.[\d.]+)\.t/
+      # Restrict versions to even numbered minor versions (except x.90+)
+      if gnome_devel_whitelist.include?(package)
+        regex ||= /#{Regexp.escape(package)}-(\d+(?:\.\d+)+)\.t/
```
Moisan: The original regex was `([\d.]+\.[\d.]+\.[\d.]+)`. Is the change to handle versions such as `1.2.3.4`? Is it possible to have something like that with Gnome versioning? Otherwise we should probably force the regex to match three digits.

livecheck/heuristic.rb
samford: `([0-8]\d*?)?` is doing two things:

1. The `[0-8]` part restricts the regex from matching x.90+ versions, which are also development versions for some GNOME formulae (e.g., `gtk+`). These versions are frequently picked up as the newest version when they should also be avoided. This _would_ also block x.900+ versions but I don't think that's worth complicating the regex further, as GNOME versions don't get that high.

2. `([0-8]\d*?)?` is used in conjunction with `[02468]` to match even minor versions > 8. Your proposed regex only matches one digit for the minor version, so version x.10 and above wouldn't match. Both of these parts work together to ensure that the _trailing_ digit (or only digit) of the minor version is even.

If we didn't have to accommodate the x.90+ development version issue, it could be simplified to `(\d+\.\d*?[02468](?:\.\d+)*?)\.t`.

livecheck/heuristic.rb
samford: My thinking here was that there could potentially be some overhead by defining the array on each iteration through the URLs, so I placed it outside of that loop to avoid any potential issue. I _did_ forget to `.freeze` this, so I need to correct that.

Is Ruby smart enough to only create the frozen array in the loop once?

livecheck/heuristic.rb
samford: This is primarily to make sure that we also match versions with only two parts (e.g., `1.2`), as this _does_ appear in some GNOME formulae (e.g., `gom`). Before this commit, `livecheck` was wrongly reporting `0.3.3` as the newest version for `gom`, when it's actually `0.4`.

However, I don't see any harm in allowing the regex to also match versions with more than three parts (as in your example). If we restricted this to only versions with 2-3 parts, we would miss any longer versions, should they occur once in a blue moon. It's unlikely to happen but it would be better to catch these versions than not.

livecheck/heuristic.rb
samford: If what I saw in profiling a simple test script (comparing creating the array outside or inside a loop) is any indicator, creating the array inside the loop will add some unnecessary overhead compared to creating it once outside the loop. This was only a quick test but it confirmed my suspicions, at least, so I think we should keep the array outside the loop.

Moisan: That makes sense, thanks for going all the way to profiling the code.

PR #311, opened 2020-03-06
Add --full-name flag

maxim-belkin:
1. Add `--full-name` flag to use fully-qualified formulae names in reports.
2. Move formatted strings (`formula_s`, `current_s`, etc) to the corresponding `if` block to avoid evaluating them when they're not needed.
3. Squash
```ruby
    needs_to_show = is_outdated || !Homebrew.args.newer_only?
    if needs_to_show
```
into
```ruby
    if is_outdated || !Homebrew.args.newer_only?
```

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -41,6 +41,10 @@ module Homebrew
     end
   end
 
+  def formula_name(formula)
+    Homebrew.args.full_name? ? formula.full_name : formula
+  end
+
   def livecheck
     livecheck_args.parse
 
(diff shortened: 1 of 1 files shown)


samford:
There are a few instances where `formula` is still used rather than `formula_name`. For example, lines 93, 99, and 115 in `print_latest_version`. There's also line 83 in `livecheck` but that's in a different scope, so `formula_name` would need to be created there to work.

At least to me, it seems like we would want that output to respect this flag as well.

maxim-belkin:
Thanks, Sam. I factored out `formula_name` into a method so that it can be used across other methods.

maxim-belkin:
ping!

samford:
I finally got around to testing this (I normally don't have other formulae taps) and it works as expected, so I'll go ahead and merge.

maxim-belkin:
Thank you, Sam!

PR #312, opened 2020-03-06
syck: add livecheckable to skip

samford:
`syck` hasn't been actively developed or maintained for quite some time. why hasn't been around since 2009 and this repo (a copy of the last version but with some additional updates since 2009) hasn't been updated since 2012. I don't imagine this situation will change, so we're better off skipping it in livecheck.

### DIFF
--- a/Livecheckables/syck.rb
+++ b/Livecheckables/syck.rb
@@ -1,0 +1,3 @@
+class Syck
+  livecheck :skip => "Not maintained"
+end


SMillerDev:
What's the benefit of skipping the check?

samford:
> What's the benefit of skipping the check?

By default, livecheck will go through the URLs in the formula (head, stable, stable mirrors, homepage) and use the heuristic to try to find versions. In the case of `syck`, it tries the GitHub repo twice: once from the stable archive URL and another from the homepage, both of which use the Git strategy (side note: we should address this in the future, sometime after `:using` is implemented). The `syck` repo doesn't tag versions and there will almost surely never be another one, so this is a waste of effort/resources. When `:skip` is used, the version checking is skipped entirely, so we avoid this issue.

Secondly, `:skip` is sort of a last resort for when we can't resolve the issue by creating a working livecheckable. Since there's no way to identify versions for this software, `livecheck` will always return with `Error: syck: Unable to get versions`. Using `:skip` and providing a very brief explanatory message of the situation says that we're aware of it but nothing can be done (unless something changes upstream). To me, this is better than failing in a way that could be construed as this formula needing a livecheckable or livecheck being broken.

It's not something that we want to be doing often but it's useful for old software that isn't maintained.

SMillerDev:
I'd say that the error describes the state of the current livecheck accurately. I'd rather have failures then something that seems to work just because we disabled it.

samford:
It’s primarily about avoiding unnecessary effort.

However, if it helps to understand it, think of it like how we treat versioned formulae. These are automatically skipped with a `versioned` notice because there’s no reason to waste effort checking for a new version (though in some cases it may be). It’s the same here; when we skip a formula, it’s because there’s no reason to waste effort checking for something that isn’t going to change.

PR #317, opened 2020-03-07
heuristic: add apache.org check

EricFromCanada:
Fetches version numbers for Apache-hosted projects at archive.apache.org by checking the directory listing for the parent directory of the path component where the version number first appears, and removed livecheckables that are covered by the new check. Also updated `zookeeper` and added an entry for `sqoop` which has an odd way of [marking prerelease versions](http://sqoop.apache.org/).

### DIFF
--- a/Livecheckables/activemq.rb
+++ b/Livecheckables/activemq.rb
@@ -1,4 +1,0 @@
-class Activemq
-  livecheck :url   => "https://activemq.apache.org/news.html",
-            :regex => /ActiveMQ ([0-9\.]+) Release/
-end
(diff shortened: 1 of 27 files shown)


samford:
I've been favoring a regex like `(\d+(?:\.\d+)+)` instead of `([0-9\.]+)`, as you're using in the heuristic changes. The latter could potentially match weird permutations like `.......`, `...1.2.3..`, etc.. In more practical terms, the most common problem that regex runs into is picking up a trailing period (e.g., `1.2.3.` from `1.2.3.tar.gz`).

The previously mentioned regex ensures that the version is something like `1.22`, `1.22.333`, etc. If you want to be able to match `1` as well, you can use something like `(\d+(?:\.\d+)*)`, of course. It's just a bit more explicit about what we're really looking for but the results are the same at the present moment for `sqoop` and `zookeeper`.

Past that, I ran this on all the formulae this applies to and it's a definite improvement. Nice work!

samford:
This just needs a rebase to resolve the `tomcat.rb` conflict and I'll merge it when I get a chance.

PR #325, opened 2020-03-08
exploitdb: delete

samford:
The existing livecheckable for `exploitdb` gives an error (`Unable to get versions`), so this removes the livecheckable to allow the heuristic to take over, as it's capable of matching versions for `exploitdb` using the Git repo tags.

### DIFF
--- a/Livecheckables/exploitdb.rb
+++ b/Livecheckables/exploitdb.rb
@@ -1,4 +1,0 @@
-class Exploitdb
-  livecheck :url   => "https://github.com/offensive-security/exploit-database/releases",
-            :regex => %r{href="/offensive-security/exploit-database/releases/tag/([0-9,\.\-]+)"}
-end


Livecheckables/exploitdb.rb
```diff
@@ -1,4 +0,0 @@
-class Exploitdb
-  livecheck :url   => "https://github.com/offensive-security/exploit-database/releases",
```
chenrui333: https://github.com/offensive-security/exploitdb/releases/tag/2020-03-07

samford: Don't worry, this PR is _removing_ the livecheckable, not updating it 😄

chenrui333: I see, I guess I am just less familiar with the flow 😄

PR #333, opened 2020-03-08
tomcat-native: add livecheckable

chenrui333:
Right now it looks like this,

```
Error: tomcat-native: Unable to get versions
```

After the change,

```
tomcat-native : 1.2.23 ==> 1.2.23
```

### DIFF
--- a/Livecheckables/tomcat-native.rb
+++ b/Livecheckables/tomcat-native.rb
@@ -1,0 +1,4 @@
+class TomcatNative
+  livecheck :url   => "https://archive.apache.org/dist/tomcat/tomcat-connectors/native/",
+            :regex => %r{href="v?(\d+(?:\.\d+)+)/"}
+end


Livecheckables/tomcat-native.rb
```diff
@@ -0,0 +1,4 @@
+class TomcatNative
+  livecheck :url   => "https://archive.apache.org/dist/tomcat/tomcat-connectors/native/",
+            :regex => %r{href="(.*?)/"}
```
samford: This regex is very loose, so if this page ever contains a linked directory that isn't a version, it could break the version matching. I see checking break for similar reasons all the time (e.g., a Git repo adding a weird tag, when before the tags were all uniform like `v.1.2.3`), so we want the regex to only match what we actually want to match.

This sometimes requires a balance between being too loose and too strict but in this case it's very easy:

```suggestion
            :regex => %r{href="(\d+(?:\.\d+)+)/"}
```

This matches the same versions as before but won't fail for the aforementioned reason.

chenrui333:
Looks like I have included some other changes.

chenrui333:
@samford PR ready for another round of review now.

PR #334, opened 2020-03-08
Update livecheck urls for tomcat formulae

chenrui333:
- use https instead of http
- update url for tomcat one with `https://archive.apache.org/dist/tomcat/`

### DIFF
--- a/Livecheckables/tomcat@7.rb
+++ b/Livecheckables/tomcat@7.rb
@@ -1,4 +1,4 @@
 class TomcatAT7
-  livecheck :url   => "http://archive.apache.org/dist/tomcat/tomcat-7/",
+  livecheck :url   => "https://archive.apache.org/dist/tomcat/tomcat-7/",
             :regex => %r{href="v(.*?)/"}
 end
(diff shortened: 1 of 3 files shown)


Livecheckables/tomcat@7.rb
```diff
@@ -1,4 +1,4 @@
 class TomcatAT7
-  livecheck :url   => "http://archive.apache.org/dist/tomcat/tomcat-7/",
+  livecheck :url   => "https://archive.apache.org/dist/tomcat/tomcat-7/",
             :regex => %r{href="v(.*?)/"}
```
samford:
```suggestion
            :regex => %r{href="v?(\d+(?:\.\d+)+)/"}
```

Livecheckables/tomcat@8.rb
```diff
@@ -1,4 +1,4 @@
 class TomcatAT8
-  livecheck :url   => "http://archive.apache.org/dist/tomcat/tomcat-8/",
+  livecheck :url   => "https://archive.apache.org/dist/tomcat/tomcat-8/",
             :regex => %r{href="v(.*?)/"}
```
samford:
```suggestion
            :regex => %r{href="v?(\d+(?:\.\d+)+)/"}
```

chenrui333:
@samford works for me. Just updated the regex to be more specific.

PR #335, opened 2020-03-08
Update .github files to ref .github ones

chenrui333:
Files are essentially the same as .github ones, https://github.com/Homebrew/.github/tree/master/.github


chenrui333:
Thanks @MikeMcQuaid!

PR #336, opened 2020-03-08
cacli: add livecheckable

chenrui333:
Add livecheckable for cacli

(note, the PR does not change the current livecheck behavior though)

### DIFF
--- a/Livecheckables/cacli.rb
+++ b/Livecheckables/cacli.rb
@@ -1,0 +1,3 @@
+class Cacli
+  livecheck :regex => /v?(\d+(?:\.\d+)+)/
+end


samford:
In the PR description, I usually try to describe the current situation of the livecheckable, why changes are needed, and what changes have been made. This allows others to understand the issue (beyond the code changes) and can help them when making changes to livecheck/livecheckables themselves.

Regarding this livecheckable, the output is the same before and after adding the livecheckable. The issue here is that version `1.04` is being interpreted as newer than `1.2.30` and this seems more like an issue with livecheck's behavior.

chenrui333:
Yeah, I was mostly thinking to surface the cacli livecheck process.

samford:
I looked deeper into how this is handled in `livecheck` and Homebrew itself. Basically everything is working as it should be and there isn't anything that can be fixed to address this.

The problem arises due to how the numeric parts of the version are parsed. In this process, the minor version of `04` becomes `4` (due to`"04".to_i`). The version string is `1.04` (what livecheck displays in the output) but the associated `Version` is effectively `1.4`. The newest version is currently `1.2.30`, so a version that's interpreted as `1.4` is going to be seen as newer than `1.2.x`. 

This is an upstream issue that we can't fix but I may work on adding something like an `:omit` feature, where we could provide tags to filter out. Usually we filter tags by carefully crafting the regex to only match what we want but it would be helpful in this case. [We could technically restrict the regex to only match single digit minor versions but this could come back to bite us later (e.g., versions `1.10.x` and above), so I would prefer to avoid it for now.]

I've added the `:omit` feature to my to-do list (after the other features), so I'm going to close this now that it's been investigated.

PR #342, opened 2020-03-08
crystal: add livecheckable

samford:
The Git tags for the `crystal` repo contain some non-release tags (e.g., `test-ci-1`, `tets-doc-4`) and this was causing the latest version to be erroneously reported as `4`. This adds a livecheckable to restrict version matching to only tags with stable versions.


issyl0:
You beat me to this! Thanks.

PR #349, opened 2020-03-09
activemq-cpp: add livecheckable

chenrui333:
# before

```
$ brew livecheck activemq-cpp
Error: activemq-cpp: Unable to get versions
```

# after

```
$ brew livecheck activemq-cpp
activemq-cpp : 3.9.5 ==> 3.9.5
```


samford:
This will be automatically addressed without needing a livecheckable when #317 is merged.

PR #356, opened 2020-03-09
ortp: update Livecheckable url

Moisan:
Update the Livecheckable with the source url used in the Formula.

### DIFF
--- a/Livecheckables/ortp.rb
+++ b/Livecheckables/ortp.rb
@@ -1,4 +1,4 @@
 class Ortp
   livecheck :url   => "https://www.linphone.org/releases/sources/ortp",
-            :regex => /href="ortp-([0-9\.]+)\.t/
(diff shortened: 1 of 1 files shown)


Livecheckables/ortp.rb
```diff
@@ -1,4 +1,4 @@
 class Ortp
-  livecheck :url   => "http://nongnu.askapache.com/linphone/ortp/sources/",
+  livecheck :url   => "https://www.linphone.org/releases/sources/ortp",
             :regex => /href="ortp-([0-9\.]+)\.t/
```
samford:
```suggestion
            :regex => /href="ortp-(\d+(?:\.\d+)+)\.t/
```

I've been trying to replace the old, potentially problematic numeric version regex (`([0-9\.]+)`) with some version of the more explicit regex (`(\d+(?:\.\d+)+)`) when I encounter it while editing livecheckables. Would you mind replacing it in this PR (as in the suggestion above)?

PR #358, opened 2020-03-09
alluxio: add livecheckable

chenrui333:
# before

```
$ brew livecheck alluxio
Error: alluxio: Unable to get versions
```

# after

There is some issue
```
$ brew livecheck alluxio
alluxio : 1.8.1 ==> 2.1.2
```

### DIFF
--- a/Livecheckables/alluxio.rb
+++ b/Livecheckables/alluxio.rb
@@ -1,0 +1,4 @@
+class Alluxio
+  livecheck :url   => "https://downloads.alluxio.io/downloads/files/",
+            :regex => %r{href="(\d+(?:\.\d+)+)/?"}
+end


samford:
For the benefit of anyone reading this after the fact, the author had asked how to match the versions on the page while omitting unstable versions like `2.2.0-RC3`. My commentary was as follows:

"Your `alluxio` livecheckable is doing a page match here, so the context is the HTML of the page. As mentioned in a previous review comment, you typically have to use something like the ending quotation mark (`"`) of the `href` attribute to restrict the matching like we do with `$` for Git tags.

In this case, you also have a trailing forward slash before the closing `"`, so we need to optionally account for that. The regex ends up looking like: `%r{href="(\d+(?:\.\d+)+)/?"}`

This only matches versions without anything after the numeric version, so the release candidate version isn't matched."

PR #360, opened 2020-03-09
gwenhywfar: update url and regex

chenrui333:
# before

```
$ brew livecheck gwenhywfar
Error: gwenhywfar: Unable to get versions
```

# after

```
$ brew livecheck gwenhywfar
gwenhywfar : 4.20.2 ==> 5.2.0
```

### DIFF
--- a/Livecheckables/gwenhywfar.rb
+++ b/Livecheckables/gwenhywfar.rb
@@ -1,4 +1,4 @@
 class Gwenhywfar
   livecheck :url   => "https://www.aquamaniac.de/rdm/projects/gwenhywfar/files",
-            :regex => /gwenhywfar-(\d+(?:\.\d+)+)\.t/
+            :regex => %r{href="[^"]+/gwenhywfar-(\d+(?:\.\d+)+)\.t}
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/gwenhywfar.rb
```diff
@@ -1,4 +1,4 @@
 class Gwenhywfar
-  livecheck :url   => "https://www.aquamaniac.de/sites/download/packages.php",
-            :regex => /gwenhywfar-([0-9\.]+)\.t/
+  livecheck :url   => "https://www.aquamaniac.de/rdm/projects/gwenhywfar/files",
+            :regex => /gwenhywfar-(\d+(?:\.\d+)+)\.t/
```
samford:
```suggestion
            :regex => %r{href="[^"]+/gwenhywfar-(\d+(?:\.\d+)+)\.t}
```

It's often better to restrict matching to the `href` attribute rather than any text on the page. As mentioned before in the context of the index pages, we never quite know what might happen with the text nodes and it's a little safer to only target the `href` attributes instead.

In the above suggestion, I decided to use the more generic `[^"]+` in place of the explicit path before the filename, since it looked like it could potentially be unstable in the future and would break matching if it changed.

PR #361, opened 2020-03-09
htpdate: update url and regex

chenrui333:
Right now it takes too much time to check on it locally.
```
Loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-livecheck/Livecheckables/htpdate.rb
Trying with url http://www.vervest.org/htp/archive/c/?C=M&O=D
Using page_match("http://www.vervest.org/htp/archive/c/?C=M&O=D", "(?-mix:href="htpdate-(\d+(?:\.\d+)+)\.t)")
Error: htpdate: execution expired
```

### DIFF
--- a/Livecheckables/htpdate.rb
+++ b/Livecheckables/htpdate.rb
@@ -1,4 +1,4 @@
 class Htpdate
-  livecheck :url   => "http://www.vervest.org/htp/download",
-            :regex => %r{href=.*?/htpdate-([0-9\.]+)\.t}
+  livecheck :url   => "http://www.vervest.org/htp/archive/c/?C=M&O=D",
+            :regex => /href="htpdate-(\d+(?:\.\d+)+)\.t/
 end


Livecheckables/htpdate.rb
```diff
@@ -1,4 +1,4 @@
 class Htpdate
-  livecheck :url   => "http://www.vervest.org/htp/download",
-            :regex => %r{href=.*?/htpdate-([0-9\.]+)\.t}
+  livecheck :url   => "http://www.vervest.org/htp/archive/c/?C=M&O=D",
+            :regex => %r{href="htpdate-(\d+(?:\.\d+)+)\.t}
```
samford:
```suggestion
            :regex => /href="htpdate-(\d+(?:\.\d+)+)\.t/
```

CI is failing due to `brew style` saying that this regex should use `//` rather than `%r{}` (since it doesn't contain a forward slash).

I would like to see if the livecheck passes on CI once you fix this.

chenrui333: 👍

PR #362, opened 2020-03-09
ibex: update regex

chenrui333:
# before

```
$ brew livecheck ibex
Error: ibex: Unable to get versions
```

# after

```
$ brew livecheck ibex
ibex : 2.8.2 ==> 2.8.7
```

### DIFF
--- a/Livecheckables/ibex.rb
+++ b/Livecheckables/ibex.rb
@@ -1,4 +1,3 @@
 class Ibex
-  livecheck :url   => "http://ibex-lib.org/download",
-            :regex => %r{href="http://www.ibex-lib.org/sites/default/files/ibex-([0-9\.]+)\.t}
+  livecheck :regex => /ibex-(\d+(?:\.\d+)+)/
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/ibex.rb
```diff
@@ -1,4 +1,3 @@
 class Ibex
-  livecheck :url   => "http://ibex-lib.org/download",
-            :regex => %r{href="http://www.ibex-lib.org/sites/default/files/ibex-([0-9\.]+)\.t}
+  livecheck :regex => /ibex-(\d+(?:\.\d+)+)/
```
samford:
```suggestion
  livecheck :regex => /^ibex-(\d+(?:\.\d+)+)$/
```

We only want to match stable versions, so you want to add a `$` to the end of the regex, to make sure we don't pick up versions with anything past the version part.

We also only want to match tags with this format, so I usually put a `^` on the front. This is in case some weird tag pops up in the future that has something before `ibex`. We wouldn't want to pick up that tag and this prevents that from happening.

Oftentimes with Git tags, you'll use `$` and `^` together like this.

chenrui333: 👍

PR #364, opened 2020-03-09
imageoptim-cli: update url and regex

chenrui333:
# before

```
$ brew livecheck imageoptim-cli --debug
...
3.0.2 => #<Version:0x00007fd335925f58 @version="3.0.2", @tokens=[#<Version::NumericToken 3>, #<Version::NumericToken 0>, #<Version::NumericToken 2>]>
issue-67 => #<Version:0x00007fd335925e68 @version="67", @tokens=[#<Version::NumericToken 67>]>
issue-67.2 => #<Version:0x00007fd335925d50 @version="67.2", @tokens=[#<Version::NumericToken 67>, #<Version::NumericToken 2>]>
imageoptim-cli (guessed) : 3.0.2 ==> 67.2
```

# after

```
$ brew livecheck imageoptim-cli --debug
...
3.0.2 => #<Version:0x00007fe309992d90 @version="3.0.2", @tokens=[#<Version::NumericToken 3>, #<Version::NumericToken 0>, #<Version::NumericToken 2>]>
imageoptim-cli : 3.0.2 ==> 3.0.2
```

### DIFF
--- a/Livecheckables/imageoptim-cli.rb
+++ b/Livecheckables/imageoptim-cli.rb
@@ -1,4 +1,3 @@
 class ImageoptimCli
-  livecheck :url   => "https://www.npmjs.com/package/imageoptim-cli",
-            :regex => %r{package__sidebarText.*?>([0-9\.]+)</p>}
+  livecheck :regex => /^v?(\d+(?:\.\d+)+)$/
 end


Livecheckables/imageoptim-cli.rb
```diff
@@ -1,4 +1,3 @@
 class ImageoptimCli
-  livecheck :url   => "https://www.npmjs.com/package/imageoptim-cli",
-            :regex => %r{package__sidebarText.*?>([0-9\.]+)</p>}
+  livecheck :regex => /^(\d+(?:\.\d+)+)$/
```
samford:
```suggestion
  livecheck :regex =>/^v?(\d+(?:\.\d+)+)$/
```

The generic regex I use for Git tags that have a format like `v1.2.3` or `1.2.3` (where that's the only text) is: `/^v?(\d+(?:\.\d+)+)$/`

The only difference is the optional `v?` on the front. Even if the tags don't ever have a `v` on the front, I still add the optional `v?` because sometimes people make a mistake and add a tag that doesn't follow suit.

If someone added a tag for `v1.2.3`, your regex wouldn't match it. I've seen this happen in upstream repos while working on livecheckables, so it's best to account for it even if it hasn't happened yet.

PR #365, opened 2020-03-09
vips: remove livecheckable

chenrui333:
# before

```
$ brew livecheck vips
Error: vips: Unable to get versions
```

# After

```
$ brew livecheck vips
vips (guessed) : 8.9.1 ==> 8.9.1
```


samford:
There's always a chance that upstream will add a weird tag to the Git repo that makes the default checking not work properly but I'm fine removing this until that point.

PR #366, opened 2020-03-09
wiredtiger: update url and regex

chenrui333:
# before

```
$ brew livecheck wiredtiger
Error: vips: Unable to get versions
```

# after

```
$ brew livecheck wiredtiger
wiredtiger : 3.2.0 ==> 3.2.1
```

### DIFF
--- a/Livecheckables/wiredtiger.rb
+++ b/Livecheckables/wiredtiger.rb
@@ -1,4 +1,4 @@
 class Wiredtiger
   livecheck :url   => "https://github.com/wiredtiger/wiredtiger.git",
-            :regex => /^mongodb-v?(\d+(?:\.\d+)+)$/
+            :regex => /^v?(\d+(?:\.\d+)+)$/
 end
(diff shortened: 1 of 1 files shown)


samford:
Looking at the tags in the Git repo, there are a number of them for other things (e.g., `mongodb-4.5.0`), so we want to restrict matching to the ones like `1.2.3`.

Instead of removing the livecheckable, modify it to only include the standard regex for Git tags: `/^v?(\d+(?:\.\d+)+)$/`

chenrui333:
@samford updated.

Livecheckables/wiredtiger.rb
```diff
@@ -1,4 +1,4 @@
 class Wiredtiger
-  livecheck :url   => "https://github.com/wiredtiger/wiredtiger/releases",
-            :regex => %r{href="/wiredtiger/wiredtiger/tree/([0-9\.]+)}
+  livecheck :url   => "https://github.com/wiredtiger/wiredtiger.git",
+            :regex => /^mongodb-v?(\d+(?:\.\d+)+)$/
```
samford: You misunderstood me. We're trying to _avoid_ the tags like `mongodb-4.5.0`, not _match_ them.

As I said in the previous review, just use the standard regex (without any modification): `/^v?(\d+(?:\.\d+)+)$/`

```suggestion
            :regex => /^v?(\d+(?:\.\d+)+)$/
```

chenrui333: I see now, works for me.

chenrui333: I see. 👍

PR #367, opened 2020-03-09
verilator: update url and regex

chenrui333:
# after

```
$ brew livecheck verilator
verilator : 4.028 ==> 4.030
```

### DIFF
--- a/Livecheckables/verilator.rb
+++ b/Livecheckables/verilator.rb
@@ -1,4 +1,4 @@
 class Verilator
-  livecheck :url   => "https://www.veripool.org/projects/verilator/wiki/Installing",
-            :regex => %r{href=".*?/verilator-([0-9\.]+)\.t}
+  livecheck :url   => "https://github.com/verilator/verilator.git",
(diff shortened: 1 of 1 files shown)


Livecheckables/verilator.rb
```diff
@@ -1,4 +1,4 @@
 class Verilator
-  livecheck :url   => "https://www.veripool.org/projects/verilator/wiki/Installing",
-            :regex => %r{href=".*?/verilator-([0-9\.]+)\.t}
+  livecheck :url   => "https://github.com/verilator/verilator.git",
+            :regex => /v(\d+(?:\.\d+)+)$/
```
samford:
```suggestion
            :regex => /^v?(\d+(?:\.\d+)+)$/
```

This is the same as before but reversed. If upstream adds a tag that doesn't include the `v` at the start, your regex wouldn't pick it up. Regardless of whether the `v` is or isn't present in the tags, I always make it optional to account for human error. Just use the standard regex for Git tags here as well: `/^v?(\d+(?:\.\d+)+)$/`

PR #368, opened 2020-03-09
telegraf: update url and regex

chenrui333:
# after the change

```
$ brew livecheck telegraf
telegraf : 1.13.4 ==> 1.13.4
```

### DIFF
--- a/Livecheckables/telegraf.rb
+++ b/Livecheckables/telegraf.rb
@@ -1,4 +1,3 @@
 class Telegraf
-  livecheck :url   => "https://portal.influxdata.com/downloads",
-            :regex => /telegraf.*?>v?([0-9\.]+)</
+  livecheck :regex => /^v?(\d+(?:\.\d+)+)$/
 end


Livecheckables/telegraf.rb
```diff
@@ -1,4 +1,3 @@
 class Telegraf
-  livecheck :url   => "https://portal.influxdata.com/downloads",
-            :regex => /telegraf.*?>v?([0-9\.]+)</
+  livecheck :regex => /(\d+(?:\.\d+)+)/
```
samford:
```suggestion
  livecheck :regex => /^v?(\d+(?:\.\d+)+)$/
```

This is picking up prerelease tags because it isn't restricted at the end. As I've been saying, if you're matching version strings like `1.2.3` or `v1.2.3` and there isn't a reason to use something different, just use the standard regex for Git tags: `/^v?(\d+(?:\.\d+)+)$/`

Be sure to check the tags beforehand, to make sure we're not omitting anything that we want to pick up, but if you're using a regex that's very similar to the standard regex, you're likely better off using the standard regex. I created it from editing quite a number of livecheckables and it has served me well.

PR #373, opened 2020-03-09
scamper: update url and regex

chenrui333:
# after the change

```
$ brew livecheck scamper
scamper : 20191102 ==> 20191102
```

### DIFF
--- a/Livecheckables/scamper.rb
+++ b/Livecheckables/scamper.rb
@@ -1,4 +1,4 @@
 class Scamper
   livecheck :url   => "https://www.caida.org/tools/measurement/scamper/code/?C=M&O=D",
-            :regex => %r{href="scamper-cvs-([0-9\.]+)\.t}
+            :regex => /href="scamper-cvs-([0-9\.]+)\.t/
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/scamper.rb
```diff
@@ -1,4 +1,4 @@
 class Scamper
-  livecheck :url   => "https://www.caida.org/tools/measurement/scamper/",
-            :regex => %r{href="code/scamper-cvs-([0-9\.]+)\.t}
+  livecheck :url   => "https://www.caida.org/tools/measurement/scamper/code/?C=M&O=D",
+            :regex => %r{href="scamper-cvs-([0-9\.]+)\.t}
```
samford:
```
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-livecheck/Livecheckables/scamper.rb ==
141
C:  3: 23: Use // around regular expression.
```

This was failing `brew style`, since it was using the `%r{}` syntax without including a `/` in the regex. You should use `//` unless the regex includes a `/`, which is when you use the `%r{}` syntax. As mentioned before, running `brew style` on your changes before submitting your PR will help you avoid these issues.

```suggestion
            :regex => /href="scamper-cvs-([0-9\.]+)\.t/
```

PR #379, opened 2020-03-09
opencv@2: update url and regex

chenrui333:
# after the change

```
$ brew livecheck opencv@2
opencv@2 : 2.4.13.7 ==> 2.4.13.7
```

### DIFF
--- a/Livecheckables/opencv@2.rb
+++ b/Livecheckables/opencv@2.rb
@@ -1,4 +1,4 @@
 class OpencvAT2
-  livecheck :url   => "https://github.com/opencv/opencv/releases",
-            :regex => %r{href="/opencv/opencv/tree/(2[0-9\.]+)"}
+  livecheck :url   => "https://github.com/opencv/opencv.git",
+            :regex => /^(2(?:\.\d+)+)$/
 end


chenrui333:
The github action diff step has some issue.
image

Livecheckables/opencv@2.rb
```diff
@@ -1,4 +1,4 @@
 class OpencvAT2
-  livecheck :url   => "https://github.com/opencv/opencv/releases",
-            :regex => %r{href="/opencv/opencv/tree/(2[0-9\.]+)"}
+  livecheck :url   => "https://github.com/opencv/opencv.git",
+            :regex => /^2.4.(\d+(?:\.\d+)+)$/
```
samford:
```suggestion
            :regex => /^(2(?:\.\d+)+)$/
```

We only want to restrict the major version here (to `2`). Though 2.4.x is the latest version, we don't want to rule out the chance of there being 2.5+ versions.

In this case, you simply substitute `2` for the first `\d+` part of the regex, as that's where the major version is.

chenrui333: good call

PR #381, opened 2020-03-09
mupdf-tools: update url

chenrui333:
# after the change

```
$ brew livecheck mupdf-tools
mupdf-tools : 1.16.1 ==> 1.16.1
```

### DIFF
--- a/Livecheckables/mupdf-tools.rb
+++ b/Livecheckables/mupdf-tools.rb
@@ -1,4 +1,4 @@
 class MupdfTools
-  livecheck :url   => "https://mupdf.com/downloads/",
-            :regex => /href="mupdf-([0-9\.]+)-source/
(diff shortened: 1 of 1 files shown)


Livecheckables/mupdf-tools.rb
```diff
@@ -1,4 +1,4 @@
 class MupdfTools
-  livecheck :url   => "https://mupdf.com/downloads/",
+  livecheck :url   => "https://mupdf.com/downloads/archive/?C=M&O=D",
             :regex => /href="mupdf-([0-9\.]+)-source/
```
samford:
```suggestion
            :regex => /href="mupdf-(\d+(?:\.\d+)+)-source/
```

Could you replace the loose version-matching regex here with the more explicit version that we normally use (as above)? I try to replace these as I find them when editing older livecheckables.

chenrui333: Sounds good

PR #383, opened 2020-03-09
libu2f-server: update url

chenrui333:
# after the change

```
$ brew livecheck libu2f-server
libu2f-server : 1.1.0 ==> 1.1.0
```

### DIFF
--- a/Livecheckables/libu2f-server.rb
+++ b/Livecheckables/libu2f-server.rb
@@ -1,4 +1,4 @@
 class Libu2fServer
-  livecheck :url   => "https://developers.yubico.com/libu2f-server/Releases",
-            :regex => /href="libu2f-server-([0-9\.]+)\.t/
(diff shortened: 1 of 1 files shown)


Livecheckables/libu2f-server.rb
```diff
@@ -1,4 +1,4 @@
 class Libu2fServer
-  livecheck :url   => "https://developers.yubico.com/libu2f-server/Releases",
+  livecheck :url   => "https://developers.yubico.com/libu2f-server/Releases/",
             :regex => /href="libu2f-server-([0-9\.]+)\.t/
```
samford:
```suggestion
            :regex => /href="libu2f-server-(\d+(?:\.\d+)+)\.t/
```

As with previous comments, while we're updating this livecheckable, could you replace the (old) loose version-matching regex with the more explicit one we use now (as above)?

chenrui333: Sounds good.

PR #390, opened 2020-03-09
john-jumbo: update url and regex

chenrui333:
# after the change

The version is 1.9.0
But it installed the jump release 1.9.0-Jumbo-1, which is different from core version, `1.9.0`

```
$ brew livecheck john-jumbo
john-jumbo : 1.9.0 ==> 1.9.0-Jumbo-1
```

Need some discussion and review in here.

### DIFF
--- a/Livecheckables/john-jumbo.rb
+++ b/Livecheckables/john-jumbo.rb
@@ -1,4 +1,4 @@
 class JohnJumbo
-  livecheck :url   => "http://www.openwall.com/john/",
-            :regex => %r{href="j/john-([a-z0-9\.\-]+)\.t}
+  livecheck :url   => "https://github.com/magnumripper/JohnTheRipper.git",
+            :regex => /^v?(\d+(?:\.\d+)+)-jumbo-\d$/i
 end


Livecheckables/john-jumbo.rb
```diff
@@ -1,4 +1,4 @@
 class JohnJumbo
-  livecheck :url   => "http://www.openwall.com/john/",
-            :regex => %r{href="j/john-([a-z0-9\.\-]+)\.t}
+  livecheck :url   => "https://github.com/magnumripper/JohnTheRipper.git",
+            :regex => /^v?(\d+(?:\.\d+)+)-Jumbo-\d$/
```
samford:
```suggestion
            :regex => /^v?(\d+(?:\.\d+)+)-jumbo-\d$/i
```

The releases here vary between using lowercase `jumbo` and title case `Jumbo`. We want to match `jumbo` without caring about the case and we can do this by adding `i` to the end of the regex, to make it case-insensitive.

With the suggested change above, livecheck matches all the jumbo versions.

chenrui333: that works for me.

PR #391, opened 2020-03-09
jbig2dec: update url and regex

chenrui333:
# after the change

The artifact is part of the release (https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/tag/gs950)

```
$ brew livecheck jbig2dec
jbig2dec : 0.17 ==> 0.17
```

### DIFF
--- a/Livecheckables/jbig2dec.rb
+++ b/Livecheckables/jbig2dec.rb
@@ -1,4 +1,4 @@
 class Jbig2dec
-  livecheck :url   => "https://artifex.com/developers-ghostscript-jbig2dec/",
-            :regex => %r{href=.*?/jbig2dec-([0-9\.]+)\.t}
+  livecheck :url   => "https://jbig2dec.com",
+            :regex => /href="[^"]*?jbig2dec-v?(\d+(?:\.\d+)+)\.t/
 end


Livecheckables/jbig2dec.rb
```diff
@@ -1,4 +1,3 @@
 class Jbig2dec
-  livecheck :url   => "https://artifex.com/developers-ghostscript-jbig2dec/",
-            :regex => %r{href=.*?/jbig2dec-([0-9\.]+)\.t}
+  livecheck :regex => %r{href=.*?/jbig2dec-v?(\d+(?:\.\d+)+)\.t}
```
samford:
```suggestion
  livecheck :url   => "https://jbig2dec.com",
            :regex => /href="[^"]*?jbig2dec-v?(\d+(?:\.\d+)+)\.t/
```

I generally only omit the URL for Git repositories, since livecheck tends to check repository URLs first and the Git strategy is the first strategy that livecheck matches. In any other situation, it's better to be explicit about the URL, to save livecheck from doing unnecessary work. In this PR, livecheck tries a number of URLs before eventually trying the homepage, so we should use the URL here.

We should also update this regex to improve it a bit. I don't like using `.*` (even lazy `.*?`) in livecheckables, since we can always be more clear about what we actually mean. As mentioned before, when working with `href` attributes, we can use `[^"]*?` (or `[^"]+`) to keep it within the bounds of the attribute. Making the leading path optional here also allows the regex to continue to match if the archive URL becomes relative in the future (i.e., `href="jbig2dec-0.17.tar.gz"`). It's unlikely to happen but it doesn't hurt to make the regex more robust against change.

PR #396, opened 2020-03-09
gcc: update url and regex

chenrui333:
With the change

```
$ brew livecheck gcc
gcc : 9.2.0 ==> 9.2.0
```

### DIFF
--- a/Livecheckables/gcc.rb
+++ b/Livecheckables/gcc.rb
@@ -1,4 +1,4 @@
 class Gcc
-  livecheck :url   => "https://gcc.gnu.org/git/gcc.git",
-            :regex => %r{^releases/gcc-([\d\.]+)$}
+  livecheck :url   => "https://ftp.gnu.org/gnu/gcc/?C=M&O=D",
+            :regex => %r{href="gcc-(\d+(?:\.\d+)+)(?:/?"|\.t)}
 end


Livecheckables/gcc.rb
```diff
@@ -1,4 +1,4 @@
 class Gcc
-  livecheck :url   => "https://gcc.gnu.org/git/gcc.git",
-            :regex => %r{^releases/gcc-([\d\.]+)$}
+  livecheck :url   => "https://ftp.gnu.org/gnu/gcc/",
```
samford:
```suggestion
  livecheck :url   => "https://ftp.gnu.org/gnu/gcc/?C=M&O=D",
```

Livecheckables/gcc.rb
```diff
@@ -1,4 +1,4 @@
...
+            :regex => /href="gcc-(\d+(?:\.\d+)+)/
```
samford:
```suggestion
            :regex => %r{href="gcc-(\d+(?:\.\d+)+)(?:/?"|\.t)}
```

Your regex matches both the folders and archive files, which is good. However, there are some archive files we don't want to match, so we need to be more explicit. In this suggestion, we're matching the folders (like `gcc-1.2.3/` but the trailing `/` is optional) and only the good archive files (like `gcc-1.2.3.tar.gz`).

chenrui333: Sounds good!

Livecheckables/gcc.rb
chenrui333: 👍

Livecheckables/gcc.rb
chenrui333: I wonder if we can introduce some test cases into livecheck repo.

So that we can actually detect these different cases.

PR #397, opened 2020-03-09
mkvtoolnix: update regex

chenrui333:
With the change
```
$ brew livecheck mkvtoolnix
mkvtoolnix : 43.0.0 ==> 44.0.0
```

Related PR, https://github.com/Homebrew/homebrew-core/pull/51341 (44.0.0 upgrade)

### DIFF
--- a/Livecheckables/mkvtoolnix.rb
+++ b/Livecheckables/mkvtoolnix.rb
@@ -1,4 +1,4 @@
 class Mkvtoolnix
   livecheck :url   => "https://mkvtoolnix.download/sources/",
-            :regex => /href="mkvtoolnix-([0-9\.]+)\.t/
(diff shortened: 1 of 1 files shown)


samford:
My review here would have completely replaced the changes in this PR, so it made more sense to create a separate PR that closes this one. Just to provide feedback, we don't want to remove the `href` attribute here. It's generally better to match the `href` attribute (rather than any text on the page), so you have to do the work of creating a regex that works in this context rather than just removing the `href` and calling it a day.

We can do this while still avoiding having to be explicit about the URL path before the filename like `href="[^"]*?`.

We also want to update the (old) loose `([0-9\.]+)` here with the more explicit `(\d+(?:\.\d+)+)`.

You can see all these changes in #414, for reference.

PR #399, opened 2020-03-09
frotz: add livecheckable

mtklr:
Current livecheck behavior

```
Error: frotz: Unable to get versions
```

After the change.

```
frotz : 2.51 ==> 2.51
```

### DIFF
--- a/Livecheckables/frotz.rb
+++ b/Livecheckables/frotz.rb
@@ -1,0 +1,4 @@
+class Frotz
+  livecheck :url   => "https://gitlab.com/DavidGriffith/frotz.git",
+            :regex => /^v?(\d+(?:\.\d+)+)$/
+end


Livecheckables/frotz.rb
```diff
@@ -0,0 +1,3 @@
+class Frotz
+  livecheck :url => "https://gitlab.com/DavidGriffith/frotz.git"
```
samford:
```suggestion
  livecheck :url   => "https://gitlab.com/DavidGriffith/frotz.git",
            :regex => /^v?(\d+(?:\.\d+)+)$/
```

This Git repo contains prerelease tags (e.g., `2.32r2`, `2.50b1`, etc.) as well as tags for other software (e.g., `dumb-2.32r1`, `xfrotz-2.32.1`). We only want to match stable releases (nothing after the numeric version, in this case) and only Frotz versions (those like `1.2.3`), so we need to restrict matching with a regex.

The standard regex I use for Git tags with versions like `1.2.3` or `v1.2.3` is `/^v?(\d+(?:\.\d+)+)$/`. This matches versions like this while also restricting matching to stable releases (due to the use of `$` at the end).

When you're working on livecheckables, you can use the `--debug` flag to see the versions that livecheck is matching. You could also use `git ls-remote --tags https://gitlab.com/DavidGriffith/frotz.git` here to see the tags in the repo, to make sure that livecheck is matching all the tags we want and omitting those we don't.

mtklr: Thanks very much for the help and explanation, I really appreciate it! I've updated the pull request with your suggested changes.
-matt

samford:
Thanks, @mtklr!

PR #424, opened 2020-03-10
digdag: add regex

samford:
The existing livecheckable for `digdag` wasn't restricting the Git tags using a regex, so a tag named `v0_10_before_merge_v0_9_38` was being listed as the newest version. This adds the standard regex we use for Git tags like `v1.2.3`, to restrict matching to these tags (and only stable versions, not `-pre1`, etc.).


samford:
Whoops, Rui already got to this in #400. Closing~~~

PR #446, opened 2020-03-12
heuristic: add bitbucket.org check

EricFromCanada:
Similar to #317, fetches version numbers for Bitbucket-hosted projects by checking either their lists of downloads or tags, and removes livecheckables that are covered by the new check. Also adds livecheckables for `amdatu-bootstrap`, `mecab-ko`, and `mecab-ko-dic` which have unusual version number schemes.


samford:
From `brew style`:

```
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-livecheck/livecheck/heuristic.rb ==
141
C:194:119: Line is too long. [119/118]
```

There have been some recent fixes that address new `brew style` changes, so you'll need to rebase this.

PR #450, opened 2020-03-14
cowsay: add livecheckable

mtklr:
before:

```
$ brew livecheck cowsay
cowsay (guessed) : 3.04 ==> 3.04
```

after:

```
$ brew livecheck cowsay
cowsay : 3.04 ==> 3.04
```

I made livechecks for some of my installed stuff - some are old/haven't been updated in a while, though, and I noticed PR #312, potentially marking this formula "skipped" because it's older and not maintained... should I just submit them and see what happens? Thanks.

### DIFF
--- a/Livecheckables/cowsay.rb
+++ b/Livecheckables/cowsay.rb
@@ -1,0 +1,4 @@
+class Cowsay
+  livecheck :url   => "https://github.com/tnalpgge/rank-amateur-cowsay.git",
+            :regex => /^cowsay-(\d+(?:\.\d+)+)$/
+end


samford:
[As mentioned in another PR](https://github.com/Homebrew/homebrew-livecheck/pull/413#issuecomment-597267716), skipping formulae is a last resort for very exceptional circumstances.

We don't skip working checks nor would we skip a non-working check for a formula if a livecheckable can be created. Not being able to check for new versions through any means is really the main factor that leads us to even _consider_ skipping a formula (not just being old and/or unmaintained).

You can see in #312 how much deliberation can go into skipping a formula that's an excellent candidate for being skipped (can't check for new versions, a new version is unlikely to be released due to not being actively developed/unmaintained, etc.).

All this having been said, livecheck is already able to check for new `cowsay` versions using the heuristic and doesn't even need a livecheckable at this point. For right now, I would only advise adding livecheckables for these formulae if the default check isn't working.

So long as we can continue checking for new versions (on GitHub or elsewhere), you have nothing to worry about. If the heuristic's check breaks in the future, for whatever reason, we'll just create a livecheckable to fix it (again, the formula won't be skipped).

If you don't mind, I'm going to close this as we're not currently focusing on adding livecheckables for formulae that are working. I'm trying to keep the number of livecheckables relatively stable for a short while, as I work on adding related features. After the features are implemented, we'll be adding quite a number of livecheckables and we may eventually add a livecheckable for this but there isn't a compelling reason right now.

mtklr:
I appreciate your explanation. I have 5 or 6 that aren't "guessed" at all, or guessed incorrectly. Just so I'm clear and don't keep bugging you - would a rule of thumb be: leave properly guessed stuff alone, and only submit PRs for incorrect or missing versions? Or should I just hold off adding any more livecheckables for now, because you're trying to work on other things instead?

thanks again,
matt

samford:
Outside of this temporary restriction, I generally prioritize 1) fixing checks where livecheck can't find a version at all, 2) fixing checks where livecheck finds the wrong version. If livecheck is guessing the version correctly, this is fine for now; it may eventually lead to the wrong version being given but there are no shortage of more pressing issues to address right now.

We have about 900 formulae where livecheck currently can't find versions and we'll need to address these either through a livecheckable or updates to the heuristic (e.g., updating or adding strategies). There are a number of others (I haven't counted) that don't have a livecheckable but need one because the default matching is giving an incorrect version (prerelease rather than stable, some other extraneous text, etc.).

I'll need to make changes to existing livecheckables for some of the features I'm working on, so having a smaller number for the moment makes this easier for me. Also, any livecheckables that we create in the interim time will likely need to be updated as features are added, so it would make for more work overall (compared to adding livecheckables afterward).

We plan to tackle these formulae after all the forthcoming livecheckable-related changes are implemented, so I've been trying to discourage adding large numbers of livecheckables for the moment. Updates to existing livecheckables are fine but I recently went through the livecheckables and nearly all of them were working properly (I addressed the ones that weren't working properly at the time).

That said, adding five or six livecheckables is fine by me, if you just want to get them out of the way.

PR #451, opened 2020-03-14
Fix newly-tweaked RuboCop style offenses in main Livecheck code

issyl0:
- [`Layout/LineLength`](https://github.com/Homebrew/brew/commit/cd3ced48e97f8fdab67b0737ba5441d4057457e9) was decreased and [`Layout/IfUnlessModifier`](https://github.com/Homebrew/brew/commit/cd3ced48e97f8fdab67b0737ba5441d4057457e9) and [`Style/ConditionalAssignment`](https://github.com/Homebrew/brew/commit/4498303f9a347184d9fc0bb6e93351bbfc4d1091) were enabled this weekend in Homebrew/brew (and hence `brew style` surfaces them here).
- This doesn't fix Livecheckables as they're annoyingly mostly _just_ above the LineLength limit. I'll leave those to people more experienced with livecheck regex syntax.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -50,7 +50,7 @@ module Homebrew
       puts $LOAD_PATH
     end
 
-    if cmd = Homebrew.args.named.first
+    if (cmd = Homebrew.args.named.first)
       require?("livecheck/commands/#{cmd}") && return
     end
 
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -102,12 +102,12 @@ module Homebrew
     has_hash_arg_with_skip =
       formula.livecheck_args.is_a?(Hash) && formula.livecheck_args.key?(:skip)
     if has_hash_arg_with_skip || formula.livecheck_args == :skip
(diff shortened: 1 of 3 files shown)


issyl0:
Is it worth overriding `Layout/LineLength` in a `.rubocop.yml` file here specifically for the `Livecheckables/` directory? :thinking:

Bo98:
I think the goal of `LineWidth` was to get it in the width that GitHub UI displays. It would be ideal to make it match, but maybe that's a lot of work to do for now.

issyl0:
Hmm. Let's merge this for now and I'll come back to the Livecheckables in a different PR.

samford:
> Is it worth overriding `Layout/LineLength` in a `.rubocop.yml` file here specifically for the `Livecheckables/` directory? 🤔

For what it's worth, livecheckables will eventually be integrated into corresponding formulae once the livecheck DSL is implemented. Looking to the future, it would be better to address the offending lines now and enforce the lint for future additions instead of overriding the cop here.

PR #452, opened 2020-03-15
Fix brew style in livecheck

Moisan:
`brew style .` was recently added to the livecheck CI but failed. This PR fixes the errors produced by `brew style .`

Notice that there is also one outstanding warning that I believe is bug we didn't spot before:
```
== cmd/livecheck.rb ==
W: 53: 12: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
```

A value is assigned to the `cmd` variable in the conditional of an `if` statement. I didn't touch it yet since the code works this way for now.

### DIFF
--- a/Livecheckables/ats2-postiats.rb
+++ b/Livecheckables/ats2-postiats.rb
@@ -1,4 +1,3 @@
 class Ats2Postiats
-  livecheck :url   => "http://www.ats-lang.org/Downloads.html",
-            :regex => %r{The current <em>stable</em> release of ATS2 is at\n<a href="http://sourceforge.net/projects/ats2-lang/download">ATS2-([0-9\.]+)</a>}
+  livecheck :regex => /ATS2-Postiats-v?(\d+(?:\.\d+)+)\.t/i
 end
--- a/Livecheckables/berkeley-db.rb
+++ b/Livecheckables/berkeley-db.rb
@@ -1,4 +1,5 @@
 class BerkeleyDb
-  livecheck :url   => "https://www.oracle.com/technetwork/database/database-technologies/berkeleydb/downloads/index.html",
-            :regex => %r{href="http://download\.oracle\.com/otn/berkeley-db/db-([0-9,\.]+)\.t}
+  livecheck :url   => "https://www.oracle.com/technetwork/database/" \
+                       "database-technologies/berkeleydb/downloads/index.html",
+            :regex => %r{href=.*?/berkeley-db/db-v?(\d+(?:\.\d+)+)\.t}
 end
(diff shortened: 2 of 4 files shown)


issyl0:
Parts of this were already done in #451.

Moisan:
> Parts of this were already done in #451.

Yep, I was a bit behind master it seems. Rebased my branch.

`capstone` is on github and the heuristic work fine on it, no need for a livecheckable.

Livecheckables/zabbix.rb
```diff
@@ -1,4 +1,5 @@
 class Zabbix
   livecheck :url   => "https://www.zabbix.com/download_sources",
-            :regex => %r{href="https://sourceforge\.net/projects/zabbix/files/ZABBIX%20Latest%20Stable/([0-9,\.]+)/zabbix-.*\.tar\.gz/download}
+            :regex => %r{href="https://sourceforge\.net/projects/zabbix/files/
+                         ZABBIX%20Latest%20Stable/([0-9,\.]+)/zabbix-.*\.tar\.gz/download}x
```
samford: We can address the line length by simplifying this livecheckable:

If we remove the URL, this defaults to the Sourceforge strategy and we can match the stable versions with the following:

```ruby
class Zabbix
  livecheck :regex => /url=.*?zabbix-(\d+(?:\.\d+)+)\.t/
end
```

If we want an explicit URL, we can use: `https://sourceforge.net/projects/zabbix/files/ZABBIX%20Latest%20Stable/`

Livecheckables/ats2-postiats.rb
```diff
@@ -1,4 +1,4 @@
 class Ats2Postiats
   livecheck :url   => "http://www.ats-lang.org/Downloads.html",
-            :regex => %r{The current <em>stable</em> release of ATS2 is at\n<a href="http://sourceforge.net/projects/ats2-lang/download">ATS2-([0-9\.]+)</a>}
+            :regex => %r{<a href="http://sourceforge.net/projects/ats2-lang/download">ATS2-([0-9\.]+)</a>}
 end
```
samford: We can address the line length by simplifying this livecheckable:

If we remove the URL, this defaults to the Sourceforge strategy and we can match the stable versions with the following:

```ruby
class Ats2Postiats
  livecheck :regex => /ATS2-Postiats-v?(\d+(?:\.\d+)+)\.t/i
end
```

If we want an explicit URL, we can use: `https://sourceforge.net/projects/ats2-lang/files/ats2-lang/`

Livecheckables/berkeley-db.rb
```diff
@@ -1,4 +1,5 @@
 class BerkeleyDb
-  livecheck :url   => "https://www.oracle.com/technetwork/database/database-technologies/berkeleydb/downloads/index.html",
+  livecheck :url   => "https://www.oracle.com/technetwork/database/" \
+                       "database-technologies/berkeleydb/downloads/index.html",
             :regex => %r{href="http://download\.oracle\.com/otn/berkeley-db/db-([0-9,\.]+)\.t}
```
samford: We can also simplify/update the regex a bit:

```ruby
            :regex => %r{href=.*?/berkeley-db/db-v?(\d+(?:\.\d+)+)\.t}
```

PR #454, opened 2020-03-15
ctags: add livecheckable

mtklr:
before:

```
Error: ctags: undefined method `[]' for nil:NilClass
```

after:

```
ctags : 5.8 ==> 5.8
```

### DIFF
--- a/Livecheckables/ctags.rb
+++ b/Livecheckables/ctags.rb
@@ -1,0 +1,4 @@
+class Ctags
+  livecheck :url   => "https://sourceforge.net/projects/ctags/",
+            :regex => /ctags-v?(\d+(?:\.\d+)+)\.t/
+end
--- a/Livecheckables/ctags.rb
+++ b/Livecheckables/ctags.rb
@@ -1,4 +1,4 @@
 class Ctags
   livecheck :url   => "https://sourceforge.net/projects/ctags/",
-            :regex => /ctags-v?(\d+(?:\.\d+)+)\.t/
+            :regex => /url=.*?ctags-v?(\d+(?:\.\d+)+)\.t/
 end


Livecheckables/ctags.rb
```diff
@@ -0,0 +1,4 @@
+class Ctags
+  livecheck :url   => "http://ctags.sourceforge.net",
+            :regex => /ctags-([0-9\.]+)\.tar\.gz/
```
samford:
```suggestion
  livecheck :url   => "https://sourceforge.net/projects/ctags/",
            :regex => /ctags-v?(\d+(?:\.\d+)+)\.t/
```

The archive file in the formula is from Sourceforge, so it's preferable for the livecheckable to check the same location.

The `[0-9\.]+` regex here is from old livecheckables and is too loose, since it doesn't adequately express our intention and can match things that we don't want to match. These days, we use some variation on the following regex: `v?(\d+(?:\.\d+)+)`

We also use `\.t` in place of the full extension for tar files, since it could change in the future (e.g., `.tgz`, `.tar.bz2`, `.tar.xz`, etc.). We have to supply _part_ of the extension to effectively restrict matching to stable versions but this is as general as we can get. You'll see `\.t` in a lot of livecheckables.

samford:
For what it's worth, the CI failure is expected here (and in the other PRs) until #452 is merged (and the PRs are rebased).

samford:
I forgot to mention that it's good to use `url=` when matching versions in the Sourceforge RSS feed (like we use `href=` on HTML pages). I pushed a commit with this change, to save you the trouble.

mtklr:
I might have messed this up by forcing a commit prior to your changes, sorry about that.

samford:
I pushed a commit to add back the change, so it's all good.

PR #455, opened 2020-03-15
fortune: add livecheckable

mtklr:
before:

```
Error: fortune: Unable to get versions
```

after:

```
fortune : 9708 ==> 9708
```

### DIFF
--- a/Livecheckables/fortune.rb
+++ b/Livecheckables/fortune.rb
@@ -1,4 +1,4 @@
 class Fortune
   livecheck :url   => "http://www.ibiblio.org/pub/linux/games/amusements/fortune/",
-            :regex => /fortune-mod-([\d\.]+)\.tar\.gz/
+            :regex => /href=.*?fortune-mod-v?(\d+(?:\.\d+)*)\.t/
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/fortune.rb
```diff
@@ -0,0 +1,4 @@
+class Fortune
+  livecheck :url   => "http://www.ibiblio.org/pub/linux/games/amusements/fortune/",
+            :regex => /fortune-mod-([\d\.]+)\.tar\.gz/
```
samford:
```suggestion
            :regex => /href=.*?fortune-mod-v?(\d+(?:\.\d+)*)\.t/
```

We prefer to match `href` attributes on HTML pages. This is especially important on index pages like these, as the file name may be truncated and give us a bad match.

As in the previous PR, we can use a variation of the standard regex for matching numeric versions here. In this case, we make additional numeric parts optional, since versions are like `9708` but could potentially be like `1.2.3` in the future.

Just for the sake of clarity, I typically use the optional `v?` in front of the numeric version, even if the file name doesn't currently use it. I've seen upstream accidentally include a version with/without a `v` that goes against their prevailing standard and we like to be resistant to these minor issues.

As before, we use `\.t` in place of the full tar extension.

PR #456, opened 2020-03-15
mame: add livecheckable

mtklr:
before:

```
mame (guessed) : 0.219 ==> 0219
```

after:

```
mame : 0.219 ==> 0.219
```

### DIFF
--- a/Livecheckables/mame.rb
+++ b/Livecheckables/mame.rb
@@ -1,0 +1,4 @@
+class Mame
+  livecheck :url   => "https://github.com/mamedev/mame/releases",
+            :regex => /MAME ([\d\.]+)/
+end


samford:
This livecheckable is working as expected for the moment (the heuristic finds the latest version using the Git tags). The issue here is that the version format from the Git tags (`0219`) doesn't match the manually-specified version format from the formula (`0.219`).

One of the features I'll be implementing will allow us to create alterations in the livecheckable (only when absolutely necessary) to massage the version from livecheck to match the version in the formula (e.g., add a period after the first digit/leading zero).

Checking for new versions in the Git tags is preferable over checking GitHub release pages (since the Git tags contain all tags but the release page only contains the latest few handful of releases on that page).

It _is_ possible to improve the livecheckable here (making the regex more restrictive, to avoid unwanted matches) but I would prefer to leave this alone for now and take care of it after I've added the alterations feature. There are a number of checks like this that I'll be fixing then and I've added this one to the list.

PR #458, opened 2020-03-15
rogue: add livecheckable

mtklr:
before:

```
rogue (guessed) : 5.4.4 ==> files
```

after:

```
rogue : 5.4.4 ==> 5.4.4
```

### DIFF
--- a/Livecheckables/rogue.rb
+++ b/Livecheckables/rogue.rb
@@ -1,0 +1,4 @@
+class Rogue
+  livecheck :url   => "https://src.fedoraproject.org/repo/pkgs/rogue/",
+            :regex => /href=.*?rogue-v?(\d+(?:\.\d+)+)\.t/
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/rogue.rb
```diff
@@ -0,0 +1,4 @@
+class Rogue
+  livecheck :url   => "https://src.fedoraproject.org/repo/pkgs/rogue/",
+            :regex => /href=.*?rogue-v?(\d+(?:\.\d+)+)\.t/
```
samford:
```suggestion
            :regex => /href=.*?rogue-?v?(\d+(?:\.\d+)+)(?:-src)?\.t/
```

The archives in this index use one of two formatting styles:

* `rogue-5.4.4.tar.gz`
* `rogue5.4.2-src.tar.gz`

There's always a chance that the next version could revert back to the older format (or use a mix of the two), so it's better to handle them both (as in the above suggestion).

PR #459, opened 2020-03-15
scummvm: update url and regex

mtklr:
before:

```
scummvm (guessed) : 2.1.1 ==> 2010-testbed
```

after:

```
scummvm : 2.1.1 ==> 2.1.1
```

### DIFF
--- a/Livecheckables/scummvm.rb
+++ b/Livecheckables/scummvm.rb
@@ -1,4 +1,4 @@
 class Scummvm
-  livecheck :url   => "https://www.scummvm.org/downloads/",
-            :regex => %r{href=".*?/scummvm-([0-9\.]+)\.t}
+  livecheck :url   => "https://github.com/scummvm/scummvm.git",
+            :regex => /^v?(\d+(?:\.\d+)+)$/
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/scummvm.rb
```diff
@@ -1,4 +1,4 @@
 class Scummvm
-  livecheck :url   => "https://www.scummvm.org/downloads/",
-            :regex => %r{href=".*?/scummvm-([0-9\.]+)\.t}
+  livecheck :url   => "https://github.com/scummvm/scummvm.git",
+            :regex => /^v?(\d+(?:\.\d+)+)$/
```
samford:
```suggestion
  livecheck :url   => "https://www.scummvm.org/frs/scummvm/",
            :regex => %r{href=.*?v?(\d+(?:\.\d+)+)/?["']}
```

The formula downloads archives from `https://www.scummvm.org/frs/scummvm/`, so we should use that index page for the livecheckable. Sometimes upstream projects will tag a version in Git before it's available in the archive format we use, so it's usually better to use an index page where the files are kept, if possible.

However, Git tags _are_ preferable when we're using a release archive from the same repository (as it's typically better than parsing release pages).

The only thing to note about the above regex is that I'm avoiding specifying `"` or `'` for the href attribute's delimiters. People should be using `"` but I've seen some upstream pages using `'`, so it's better to be ambiguous if we can (as it may be more resistant to upstream changes over time.

PR #464, opened 2020-03-18
wp-cli: add livecheckable

samford:
This adds a livecheckable for `wp-cli` that makes it only check the release that's marked as "latest" on GitHub. There's a `2.4.1` version on GitHub (which livecheck is reporting as newest) but the ["latest" release](https://github.com/wp-cli/wp-cli/releases/latest) is currently 2.4.0 and [the README on GitHub](https://github.com/wp-cli/wp-cli/blob/master/README.md#wp-cli) corroborates that this is the latest stable version.

For what it's worth, to make this work, it was necessary to exclude URLs that contain `/latest` from using the Git strategy by default, otherwise we're unable to do a page match. I tested the other formulae containing `/latest` URLs (`kotlin`, `prometheus`, `pyenv-virtualenv`, `sysdig`) and they continue to work fine.

### DIFF
--- a/Livecheckables/wp-cli.rb
+++ b/Livecheckables/wp-cli.rb
@@ -1,0 +1,4 @@
+class WpCli
+  livecheck :url   => "https://github.com/wp-cli/wp-cli/releases/latest",
+            :regex => /href=".*?v?(\d+(?:\.\d+)+)"/
+end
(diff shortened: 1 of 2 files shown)


samford:
I originally used the following for this livecheckable:

```ruby
class WpCli
  livecheck :url   => "https://api.github.com/repos/wp-cli/wp-cli/releases/latest",
            :regex => %r{"[^"]+/wp-cli-v?(\d+(?:\.\d+)+).phar"}
end
```

This worked locally but the `brew livecheck` step of the CI checks was running into a `403 Forbidden` error when trying to access the URL. I modified the livecheckable to use the "releases/latest" URL on github.com, which provides the URL to the latest release.

PR #466, opened 2020-03-20
Factor out special cases into arrays

maxim-belkin:
Factoring out github/sourceforge/gnu special cases into array variables to standardize the approach. Should provide minor speed improvements.

### DIFF
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -11,6 +11,44 @@ def version_heuristic(livecheckable, urls, regex = nil)
     "libepoxy",
   ].freeze
 
+  github_special_cases = %w[
+    api.github.com
+    /latest
+    menhir
+    mednafen
+    camlp5
+    kotlin
+    osrm-backend
+    prometheus
+    pyenv-virtualenv
+    sysdig
+    shairport-sync
+    yuicompressor
+  ].freeze
+
+  sourceforge_special_cases = %w[
+    mikmod
+    log4cpp
+    exiftool
+    libwps
+    gsmartcontrol
+    e2fsprogs
+    potrace
+    remake
+    /avf/
+    /bashdb/
+    /netpbm/
+    opencore-amr
+  ].freeze
+
+  gnu_special_cases = %w[
+    kawa
+    lzip
+    numdiff
+    icoutils
+    dvdrtools
+  ].freeze
+
   urls.each do |url|
     # Skip Gists until/unless we create a method of identifying revisions
     next if url.include?("gist.github.com")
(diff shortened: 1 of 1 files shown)


samford:
> Should provide minor speed improvements.

I did some quick benchmarking of this in isolated test scripts and the proposed array setup seemed just a _little_ bit slower than the current setup for me. It could just be my local system or the way I was benchmarking this, so it would be good for others to run their own tests.

Would you have any interest in benchmarking this (and providing your test code) to confirm any performance changes?

maxim-belkin:
Performance was not my main goal, tbh -- standardization and code clarity were. Do you see performance as an important part of this contribution? I can do some benchmarks if you do.

samford:
I only mentioned performance due to the comment about speed improvements and my benchmarks contradicting the idea. Though it's always nice to keep performance in mind and to make things more efficient when we can, the negative difference here is relatively minor in practice, so I'm okay with it.

Performance aside, I do agree this makes it easier to visually parse the heuristic's big `if`/`elsif` block. For what it's worth, I'll be cleaning up `version_heuristic` a bit more as part of implementing the `:using` feature, so it should become even easier to read in the future (beyond the improvement here).

I'm going to leave this PR open for a little bit to see if anyone else has any comments and I'll merge it after a few days. Thanks, Maxim!

maxim-belkin:
Thanks, Sam. In my experience, it is better to request reviews  -- otherwise PRs tend to stay unnoticed/unreviewed.

samford:
I agree but I'm not sure that any other maintainers are 1) familiar with the codebase, 2) interested in reviewing the more complicated PRs (not simple livecheckable modifications), and 3) available for review. I've sort of become the de facto livecheck person for now, since I'm interested in seeing the project move forward (granted, I don't have as much time to work on it as I would like).

Before I started working on livecheck, Thierry seemed to be the main person managing issues and PRs but he hasn't been as active lately (which I certainly understand). I've tried requesting reviews on livecheck PRs in the past but I didn't get any responses, so I've given up on it until the situation changes (as I don't want to bother people who may not have the time to work on reviews).

However, I do need to find some kind of balance between promptly merging PRs and indefinitely leaving them open for discussion. Simply leaving them open for discussion without requesting reviews doesn't seem to be a working strategy but I'm wary of merging more complex PRs quickly, as I don't wish to become the sole arbiter (though it sometimes works like that these days).

I think maybe merging good PRs after a week (being mindful of the weekend) might be an okay approach. I'm going to go ahead and merge this, since it's pretty straightforward and there's a related change I would like to make.

PR #494, opened 2020-03-28
apache-brooklyn-cli: add livecheckable

chenrui333:
# before the change

```
$ brew livecheck apache-brooklyn-cli
apache-brooklyn-cli (guessed) : 1.0.0 ==> 1.0.0-M1
```

<details>
<summary>brew livecheck apache-brooklyn-cli --debug</summary>

```
Trying with url https://github.com/apache/brooklyn-client/archive/rel/apache-brooklyn-1.0.0.tar.gz
Possible git repo detected at https://github.com/apache/brooklyn-client.git
rel/apache-brooklyn-0.10.0 => #<Version:0x00007f936b1e97f8 @version="0.10.0", @tokens=[#<Version::NumericToken 0>, #<Version::NumericToken 10>, #<Version::NumericToken 0>]>
rel/apache-brooklyn-0.11.0 => #<Version:0x00007f936b1e92a8 @version="0.11.0", @tokens=[#<Version::NumericToken 0>, #<Version::NumericToken 11>, #<Version::NumericToken 0>]>
rel/apache-brooklyn-0.12.0 => #<Version:0x00007f936b1e8f60 @version="0.12.0", @tokens=[#<Version::NumericToken 0>, #<Version::NumericToken 12>, #<Version::NumericToken 0>]>
rel/apache-brooklyn-0.9.0 => #<Version:0x00007f936b1e8ba0 @version="0.9.0", @tokens=[#<Version::NumericToken 0>, #<Version::NumericToken 9>, #<Version::NumericToken 0>]>
rel/apache-brooklyn-1.0.0 => #<Version:0x00007f936b1e88f8 @version="1.0.0", @tokens=[#<Version::NumericToken 1>, #<Version::NumericToken 0>, #<Version::NumericToken 0>]>
rel/apache-brooklyn-1.0.0-M1 => #<Version:0x00007f936b1e87e0 @version="1.0.0-M1", @tokens=[#<Version::NumericToken 1>, #<Version::NumericToken 0>, #<Version::NumericToken 0>, #<Version::StringToken "M1">]>
apache-brooklyn-cli (guessed) : 1.0.0 ==> 1.0.0-M1
```

</details>

# after the change

```
$ brew livecheck apache-brooklyn-cli
apache-brooklyn-cli : 1.0.0 ==> 1.0.0
```

### DIFF
--- a/Livecheckables/apache-brooklyn-cli.rb
+++ b/Livecheckables/apache-brooklyn-cli.rb
@@ -1,0 +1,4 @@
+class ApacheBrooklynCli
+  livecheck :url   => "https://github.com/apache/brooklyn-client.git",
+            :regex => %r{^rel\/apache-brooklyn-?(\d+(?:\.\d+)+)$}i
+end
--- a/Livecheckables/apache-brooklyn-cli.rb
+++ b/Livecheckables/apache-brooklyn-cli.rb
@@ -1,4 +1,4 @@
 class ApacheBrooklynCli
   livecheck :url   => "https://github.com/apache/brooklyn-client.git",
-            :regex => %r{^rel\/apache-brooklyn-?(\d+(?:\.\d+)+)$}i
+            :regex => %r{^(?:rel/)?apache-brooklyn-(\d+(?:\.\d+)+)$}i
 end


Livecheckables/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,4 @@
+class ApacheBrooklynCli
+  livecheck :url   => "https://github.com/apache/brooklyn-client.git",
+            :regex => %r{^rel\/apache-brooklyn-?(\d+(?:\.\d+)+)$}i
```
samford:
```suggestion
            :regex => %r{^(?:rel/)?apache-brooklyn-(\d+(?:\.\d+)+)$}i
```

The `%r{}` syntax is recommended by `brew style` when a `/` is present in the regex because using the `%r{}` syntax allows us to use forward slashes without having to escape them. As such, we can unescape the forward slash here now that the syntax is updated.

Some of the (early?) tags don't have the leading `rel/` text, so we may want to make it optional here in case they move away from this scheme or forget to add it to a release. We're restricting matching to stable versions, so we don't have to worry about picking up unstable versions if they end up publishing pre-release tags without the preceding `rel/`.

The hyphen before the numeric version seems to always be present in the tags, so it's probably safe to drop the `?` in `apache-brooklyn-?`.

Issue #267, opened 2020-03-01
livecheck does not check the open PR


chenrui333:
Currently, livecheck does not report if there is open PR for the possible upgrade. It would be nice to have it integrated with the reporting.

Moisan:
Note that `brew bump-formula-pr` fails if there is an existing PR. We could probably use that check here.

samford:
I imagine this is something that we would want to put behind an option flag, as it may be a relatively intensive check to make and could also potentially bloat livecheck's output (depending on how this information is presented).

That is to say, we wouldn't want this to pull PR data with each individual livecheck, as this would be wasteful in a multiple-formula run. Rather, it would be better to simply collect the data on all open PRs once at the start of the run and search that data on each individual livecheck.

Looking to the future, I've been working on an idea that would effectively provide this information to livecheck and other tools, without requiring the tool itself to search for PRs manually. I'm still working on fleshing out this idea more and I would like to create a dumb proof of concept first (i.e., not including all the technical underpinnings that would make it possible in production) before presenting it for discussion. I wanted to mention this at least (albeit vaguely), as it may be better to hold off on this feature unless it ends up being a very easy addition.

chenrui333:
> Note that brew bump-formula-pr fails if there is an existing PR. We could probably use that check here.

Agree, we should be able to leverage the common logic for the similar operations.

MikeMcQuaid:
Closing for now but can be moved or reopened in Homebrew/brew if it applies to the new `brew livecheck` there.

PR #499, opened 2020-04-02
livecheck: add possibility to check open PRs

dawidd6:
Closes: #267

PR adding an option to print open PRs for checked formulae.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -37,12 +37,14 @@ module Homebrew
              description: "Check all available formulae."
       switch "--newer-only",
              description: "Show the latest version only if it's newer than the formula."
+      switch "--check-pulls",
+             description: "Check open pull requests."
       conflicts "--tap=", "--all", "--installed"
     end
   end
 
   def formula_name(formula)
-    Homebrew.args.full_name? ? formula.full_name : formula
+    Homebrew.args.full_name? ? formula.full_name : formula.name
   end
 
   def livecheck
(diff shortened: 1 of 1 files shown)


samford:
I'll try to get around to testing this out and reviewing it sometime in the next few days but I wanted to make sure to find the time to give it a good look.

cmd/livecheck.rb
```diff
@@ -83,18 +85,79 @@ def livecheck
...
+      end
+
+      repos = formulae.map do |formula|
+        formula_repo formula
+      end
+
+      repos.uniq!
+
+      repos.each do |repo|
+        pulls = GitHub.search("issues", type: "pr", state: "open", repo: repo)
```
samford, line 112: There may be some value in referring to the PRs as `pull_requests` rather than `pulls`. The name is a little ambiguous to me and I think it's best to just be explicit. `pull_requests` would also bring this in line with the code in `bump_formula-pr.rb`, which does something similar.

cmd/livecheck.rb
```diff
@@ -37,12 +37,14 @@ def livecheck_args
              description: "Check all available formulae."
       switch "--newer-only",
              description: "Show the latest version only if it's newer than the formula."
+      switch "--check-pulls",
```
samford, line 40: With the other naming comments in mind, maybe `--check-prs` instead?

cmd/livecheck.rb
```diff
@@ -83,18 +85,79 @@ def livecheck
...
+        pulls.each do |pull|
+          formulae.each do |formula|
+            next unless formula_repo(formula) == repo
+            next unless pull["title"].include?(formula.name)
+
+            if Homebrew.args.json?
+              formulae_info_checked.each_index do |ind|
+                next unless formulae_info_checked[ind]["formula"] == formula_name(formula)
+
+                formulae_info_checked[ind]["pull"] = pull["title"]
```
samford, line 122: This is perhaps a little bikeshed-y but I usually name the index variable `i`, to keep it brief and align it with the longstanding use of `i` as the index variable in `for` loops, etc. in other languages. `i` is used for the `each_index` variable name in Homebrew/brew, so I think we should follow suit.

---

This only seems to allow for finding one open PR containing the formula name in the title. If there's more than one related PR open, this would only return the last one it finds, since any previous value would be overwritten. We want to find all the open PRs for a formula, so we should be using something like an array here and pushing the matching PRs to it.

We may also want to rename the variable to make it explicit, as in the previous comment. I think `pull_requests` readily explains the purpose whereas `pulls` would be vague to me if I didn't already know the purpose.

`bump-formula-pr` also outputs both the PR title and URL, so we probably want to do that, too. If we make the pushed value an object with `title` and `url` keys, that could be more useful than just pushing the PR title with no accompanying URL.

With all this in mind, it could be like:

```suggestion
              formulae_info_checked.each_index do |i|
                next unless formulae_info_checked[i]["formula"] == formula_name(formula)

                formulae_info_checked[i]["pull_requests"].push({ "title" => pr["title"], "url" => pr["html_url"] })
```

dawidd6:
Thanks for detailed review @samford.
Just wanted to say, if someone wants to continue the work on this, then please go ahead, even push to this branch.

PR #500, opened 2020-04-03
spice-gtk: update regex

EricFromCanada:
Before:
```
spice-gtk : 0.37 ==> 0.37
```
After:
```
spice-gtk : 0.37 ==> 0.38
```

### DIFF
--- a/Livecheckables/spice-gtk.rb
+++ b/Livecheckables/spice-gtk.rb
@@ -1,4 +1,4 @@
 class SpiceGtk
   livecheck :url   => "https://www.spice-space.org/download/gtk/",
-            :regex => /spice-gtk-([\d.]+)\.tar\.xz/
+            :regex => /href=.*?spice-gtk-(\d+(?:\.\d+)+)\.t/
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/spice-gtk.rb
```diff
@@ -1,4 +1,4 @@
 class SpiceGtk
   livecheck :url   => "https://www.spice-space.org/download/gtk/",
-            :regex => /spice-gtk-([\d.]+)\.tar\.bz2/
+            :regex => /spice-gtk-([\d.]+)\.tar\.xz/
```
samford, line 3:
```suggestion
            :regex => /href=.*?spice-gtk-(\d+(?:\.\d+)+)\.t/
```

Typically the way livecheckables deal with the various `tar` extensions is to simply use `\.t`in the regex.

Past that, we can improve this regex a bit, while we're at it. With index pages, it's best to only match the file names in the `href` attribute, since the file name in text nodes can be truncated. Additionally, the capture group for the version string is basically matching anything (but doesn't intend to) since the `.` is unescaped, so we may as well replace it with the prevailing regex for matching numeric versions.

PR #527, opened 2020-04-12
nuget: update url and regex

samford:
#519 updated the livecheckable for `nuget` to properly identify the newest version, rather than just finding the version that appears to be newest (w.r.t., [a discussion in a `nuget` update PR in homebrew-core](https://github.com/Homebrew/homebrew-core/pull/52939/files#r407134065)).

This modifies the previous method to look for `ReleasedAndBlessed` versions in https://dist.nuget.org/tools.json instead, which may be more reliable in the long run (as the previous check would break if the "recommended latest" text ever changes to something else). As a nice side effect, this will pick up all the previous `ReleasedAndBlessed` versions rather than just one version.

### DIFF
--- a/Livecheckables/nuget.rb
+++ b/Livecheckables/nuget.rb
@@ -1,4 +1,4 @@
 class Nuget
-  livecheck :url   => "https://dist.nuget.org/index.json",
-            :regex => /"displayName":\s*?"nuget.exe - recommended latest",\s*?"version":\s*?"v?(\d+(?:\.\d+)+)"/i
+  livecheck :url   => "https://dist.nuget.org/tools.json",
+            :regex => %r{"url":\s*?"[^"]+/v?(\d+(?:\.\d+)+)/nuget\.exe",\s*?"stage":\s*?"ReleasedAndBlessed"}i
 end


Livecheckables/nuget.rb
```diff
@@ -1,4 +1,4 @@
 class Nuget
-  livecheck :url   => "https://dist.nuget.org/index.json",
-            :regex => /"displayName":\s*?"nuget.exe - recommended latest",\s*?"version":\s*?"v?(\d+(?:\.\d+)+)"/i
```
chenrui333, line 3: IMO, index.json and tools.json are having the same infromation?

samford: It's similar information but the main difference is that the objects have a `stage` property which tells you if a new version is simply `Released` (not what we want to use) or `ReleasedAndBlessed` (the versions we're interested in). For example, `5.5.1` is `Released` whereas `5.4.0` is `ReleasedAndBlessed`. [There's also an `EarlyAccessPreview` stage but clearly we don't want to use those.]

As mentioned above, with `index.json` we rely on the `nuget.exe - recommended latest` text always remaining the same and we only match the latest release this way. If the text changes, even slightly, the check will break and we won't get the latest version.

With `tools.json`, we're able to match all the versions that have been marked as `ReleasedAndBlessed` and livecheck will determine which one is newest. This gives us more meaningful output when running `brew livecheck --debug nuget` and is arguably less brittle than the `index.json` method.

chenrui333: > arguably less brittle than the `index.json` method.

Indeed, I can actually imagine both of these generated information are from the same source though, but the type key `ReleasedAndBlessed` is indeed short and notable.

samford: They're from the same source but the way the information is presented impacts how we match versions in the livecheckable and what versions the check is able to find. I think it's worthwhile to use `tools.json` instead, for the reasons I've previously mentioned. Have I convinced you?

chenrui333: I am OK with the change. 🚀

PR #528, opened 2020-04-12
node@12: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck node@12
node@12 : versioned
```

### after the change

```
$ brew livecheck node@12
node@12 : 12.16.2 ==> 12.16.2
```

----
relates to 6f402de58a8e468783cbfcd5962826488fc4511f (node@10 livecheckable)


samford:
I'll review this after #536 is merged and suggest an alternative version of this livecheckable that uses the version restriction feature, as it'll allow us to use the Git repo tags (like the check for the `node` formula) without needing to restrict version matching with the regex.

PR #534, opened 2020-04-12
onioncat: update livecheck to check against github

chenrui333:
### before the change

```
$ brew livecheck onioncat
onioncat : 0.2.8 ==> 0.3.3a
```

### after the change

```
$ brew livecheck onioncat
onioncat : 0.2.8 ==> 0.2.8
```

### DIFF
--- a/Livecheckables/onioncat.rb
+++ b/Livecheckables/onioncat.rb
@@ -1,4 +1,6 @@
 class Onioncat
-  livecheck :url   => "https://www.cypherpunk.at/ocat/download/Source/current/",
-            :regex => /href="onioncat-([0-9\.]+.*?)\.t/
+  # Per https://www.onioncat.org/download/
+  # The source code of OnionCat is managed on Github since 9th of January 2018.
+  livecheck :url   => "https://github.com/rahra/onioncat.git",
+            :regex => /^v?(\d+(?:\.\d+)+)$/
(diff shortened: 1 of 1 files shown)


Livecheckables/onioncat.rb
```diff
@@ -1,4 +1,6 @@
 class Onioncat
-  livecheck :url   => "https://www.cypherpunk.at/ocat/download/Source/current/",
-            :regex => /href="onioncat-([0-9\.]+.*?)\.t/
+  # Per https://www.onioncat.org/download/
+  # The source code of OnionCat is managed on Github since 9th of January 2018.
+  livecheck :url   => "https://github.com/rahra/onioncat.git",
+            :regex => /^v?(\d+(?:\.\d+)+)$/
```
samford, line 5:
```suggestion
  livecheck :url   => "https://www.cypherpunk.at/ocat/download/Source/stable/",
            :regex => /href=.+?onioncat-v?(\d+(?:\.\d+)+)\.t/
```

The formula uses the archives at cypherpunk.at/ocat and we can address this issue by checking the archive page for `stable` instead of `current`, so there's no need to use the Git repo here.

The suggestion above simply modifies the existing livecheckable to use the `stable` URL and updates the regex to bring it in line with current standards.

PR #553, opened 2020-04-19
secure urls (http -> https)

chenrui333:
use https endpoint instead of http endpoint for checks.

### DIFF
--- a/Livecheckables/druid.rb
+++ b/Livecheckables/druid.rb
@@ -1,4 +1,4 @@
 class Druid
-  livecheck :url   => "http://druid.io/downloads.html",
+  livecheck :url   => "https://druid.io/downloads.html",
             :regex => /href=".*druid-([0-9,\.]+)-bin\.t/
 end
--- a/Livecheckables/druid.rb
+++ b/Livecheckables/druid.rb
@@ -1,4 +1,4 @@
 class Druid
-  livecheck :url   => "https://druid.io/downloads.html",
+  livecheck :url   => "https://druid.apache.org/downloads.html",
             :regex => /href=".*druid-([0-9,\.]+)-bin\.t/
 end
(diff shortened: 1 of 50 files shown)


Livecheckables/druid.rb
```diff
@@ -1,4 +1,4 @@
 class Druid
-  livecheck :url   => "http://druid.io/downloads.html",
+  livecheck :url   => "https://druid.io/downloads.html",
```
samford, line 2:
```suggestion
  livecheck :url   => "https://druid.apache.org/downloads.html",
```

This redirected when I tested it, so we should update the URL to avoid the redirection.

Livecheckables/nspr.rb
```diff
@@ -1,4 +1,4 @@
 class Nspr
-  livecheck :url   => "http://ftp.mozilla.org/pub/mozilla.org/nspr/releases/",
+  livecheck :url   => "https://ftp.mozilla.org/pub/mozilla.org/nspr/releases/",
```
samford, line 2:
```suggestion
  livecheck :url   => "https://ftp.mozilla.org/pub/nspr/releases/",
```

This redirects to https://ftp.mozilla.org/pub/nspr/releases/, so we should update the URL to avoid the redirection.

Livecheckables/paperkey.rb
```diff
@@ -1,4 +1,4 @@
 class Paperkey
-  livecheck :url   => "http://www.jabberwocky.com/software/paperkey",
+  livecheck :url   => "https://www.jabberwocky.com/software/paperkey",
```
samford, line 2:
```suggestion
  livecheck :url   => "https://www.jabberwocky.com/software/paperkey/",
```

This redirects to the same URL with a trailing `/`, so we should update the URL to avoid the redirection.

Livecheckables/qd.rb
```diff
@@ -1,4 +1,4 @@
 class Qd
-  livecheck :url   => "http://crd-legacy.lbl.gov/~dhbailey/mpdist/",
+  livecheck :url   => "https://crd-legacy.lbl.gov/~dhbailey/mpdist/",
```
samford, line 2:
```suggestion
  livecheck :url   => "https://www.davidhbailey.com/dhbsoftware/",
```

This redirects to https://www.davidhbailey.com/dhbsoftware/, so we should update the URL to avoid the redirection.

samford:
I was working on seeing if there were any other livecheckables that this would apply to but I'll just take care of the ones I found in a follow-up PR.

PR #555, opened 2020-04-24
Add support for livecheck DSL

nandahkrishna:
This PR aims to add fallback code to livecheck, in order to handle the livecheck Formula DSL (Homebrew/brew#7027) for `skip`.

With this addition, any `skip` added to `some-formula.rb` in homebrew/core as part of the livecheck DSL will be handled properly. See Homebrew/brew#7179 for more details.

Example:
some-formula.rb in homebrew/core:
```ruby
...
  livecheck do
    skip "Not maintained"
  end
...
```
Output of `brew livecheck some-formula`:
```bash
some-formula: skipped - Not maintained
```

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -109,14 +109,11 @@ module Homebrew
       return
     end
 
-    has_hash_arg_with_skip =
-      formula.livecheck_args.is_a?(Hash) && formula.livecheck_args.key?(:skip)
     is_gist = formula.stable.url.include?("gist.github.com")
-    if has_hash_arg_with_skip || formula.livecheck_args == :skip || is_gist
-      skip_msg = if has_hash_arg_with_skip &&
-                    formula.livecheck_args[:skip].is_a?(String) &&
-                    !formula.livecheck_args[:skip].empty?
-        " - #{formula.livecheck_args[:skip]}"
+    if formula.livecheck.skip? || is_gist
+      skip_msg = if formula.livecheck.skip_msg.is_a?(String) &&
+                    !formula.livecheck.skip_msg.blank?
+        " - #{formula.livecheck.skip_msg}"
       elsif is_gist
         " - Stable URL is a GitHub Gist"
       else
--- a/livecheck/extend/formula.rb
+++ b/livecheck/extend/formula.rb
@@ -1,21 +1,13 @@
 require "formula"
 
 class Formula
+  delegate :livecheck => :"self.class"
+
   def latest
     self.class.latest
   end
 
-  def livecheck_args
-    self.class.livecheck_args
-  end
-
-  def livecheckable?
-    self.class.livecheckable == true
-  end
-
   class << self
-    attr_reader :livecheck_args, :livecheckable
-
     def all_urls
       urls = []
       urls << head.url if head
@@ -29,28 +21,23 @@ class Formula
     end
 
     def livecheck(arg = {}, &block)
-      @livecheck_args = block || arg
+      @livecheck ||= Livecheck.new
+      return @livecheck if livecheckable? || (!block_given? && arg.empty?)
+
       @livecheckable = true
+      if block_given?
+        @livecheck.instance_eval(&block)
+      else
+        arg.each do |key, value|
+          @livecheck.send(key.to_s, value)
+        end
+      end
     end
 
     def _latest
-      @livecheck_args ||= {}
-      version_s =
-        if @livecheck_args.is_a? Proc
-          result = @livecheck_args.call
-          if result.is_a? Array
-            result.map { |s| Version.new(s) }.max
-          else
-            result
-          end
-        else
-          urls = [@livecheck_args[:url]] if @livecheck_args[:url]
-          urls ||= all_urls
-          regex = @livecheck_args[:regex]
-
-          version_heuristic(livecheckable, urls, regex)
-        end
-
+      urls = [@livecheck.url] if @livecheck.url.is_a?(String) && !@livecheck.url.blank?
+      urls ||= all_urls
+      version_s = version_heuristic(livecheckable?, urls, @livecheck.regex)
       Version.new(version_s)
     end
 


cmd/livecheck.rb
```diff
@@ -122,7 +122,7 @@ def print_latest_version(formula)
         ""
       end
 
-      puts "#{Tty.red}#{formula_name(formula)}#{Tty.reset} : skipped#{skip_msg}" unless Homebrew.args.quiet?
+      puts "#{Tty.red}#{formula_name(formula)}#{Tty.reset}: skipped#{skip_msg}" unless Homebrew.args.quiet?
```
samford, line 125: It's standard behavior for the output to have a space both before and after the colon (there are currently seven instances of this in `livecheck.rb`), so this is intentional and shouldn't be removed.

cmd/livecheck.rb
```diff
@@ -132,6 +132,11 @@ def print_latest_version(formula)
       latest = Version.new(m[1])
     end
 
+    if latest.to_s.include? "skipped"
```
samford, line 135: The main issue I have with handling `skip` in this manner is that we shouldn't even be getting to the point where `latest` is executed. The entire point of the skip feature is that livecheck exits early with the skip message before `latest` is even called. The current `skip` implementation takes care of this earlier in `print_latest_version` and we should simply extend that logic to work with the livecheck DSL's `skip`. [I've already created a tentative implementation of exactly this and will post code in the main review comment.]

Additionally, with the way this PR currently handles things, the `latest` value is the skip message and this feels wrong from a semantic standpoint (i.e., `latest` should be the latest version, not the skip message).

Besides all this, there's a remote chance that a future formula will contain "skipped" and this would fail in that case.

cmd/livecheck.rb
nandahkrishna: Whoops, that was an accident, sorry 😅

cmd/livecheck.rb
nandahkrishna: Agreed. This particular implementation was written in order to "just work" based on the current constraints (without modifying the other code too much), and is certainly not elegant.

nandahkrishna:
I'll pause on this PR for now, until the changes you mentioned are done. Thanks!

nandahkrishna:
@samford Just thought this PR could be a place to discuss my implementation of handling the block in Homebrew/homebrew-livecheck.

My latest commit reflects a step in that direction, when I was getting familiar with the codebase here and trying out things. It is by no means complete, but it does seem to work for `aacgain` and `imap-uw` for which I've raised PRs in Homebrew/homebrew-core, and additionally doesn't seem to tamper with existing Livecheckables.

nandahkrishna:
Thanks for the comments. Just thinking aloud here to log my thoughts.

### About not duplicating logic

Definitely, our aim would be to handle both the Livecheckable and the block without too many different conditions and branches. As we're looking at a transition to the blocks in favour of the Livecheckables, would it be better to rewrite code to work based on arguments retrieved from the `Livecheck` class (more specifically, the output of `#to_hash`)? And temporarily, for Livecheckable support, convert the format of `@livecheck_args` to be consistent with the former. Wouldn't be too much of a difference, I think.

### What if both a block and a livecheckable exist?

I think that's been handled properly as of now (`block || args`) - possible rewrite required for better handling -  and I tested it by adding relevant debug statements here and there. The block takes priority over the Livecheckable, and I think that should be the case - after all we're eschewing the Livecheckables in favour of the block. And as we're pretty much just "translating" the Livecheckable into the block, there shouldn't be an issue with this.

Disclaimer: these are just thoughts and I haven't finished thinking about the problems that would arise. Awaiting feedback, if you have some points or questions to add, while I try to implement these.

Thanks!

samford:
> The block takes priority over the Livecheckable, and I think that should be the case

Yes, exactly this 👍

> "translating" the Livecheckable into the block

I was hoping you would discover this as well. As you mentioned, this setup would allow us to update all the logic in livecheck that accesses the livecheckable data (e.g., `formula.livecheck_args[:url]`) to the new format (e.g., `formula.livecheck.url`), regardless of whether the source is a block in a formula or a hash from a livecheckable file. When it's time to remove support for older livecheckables, we should only need to do some small removals in one place (not all throughout livecheck).

Feel free to push any changes you would like to be reviewed as you work through this and let me know if you run into any issues or have any questions along the way.

nandahkrishna:
@samford, I got some work done on this. Tested it out locally on `aacgain` and `imap-uw` (both as Livecheckables and DSL blocks), and it seems to work. Looking forward to your suggestions!

SMillerDev:
> Also, we would have to move the _latest method to formula.rb in Homebrew/brew during the transition, right? If so, would it be wise to raise a PR for that now or later on?

Without comment on this specific situation. A lot of smaller updates over time are way easier to review and get right then one larger one.

nandahkrishna:
Understood. Was just a _little_ too enthusiastic about getting work done, haha 😅

SMillerDev:
> Understood. Was just a _little_ too enthusiastic about getting work done, haha 😅

Is there such a thing?

Hype overload

I'm just suggesting you send a small pull request when you have it ready instead of bundling it all up.

livecheck/extend/formula.rb
```diff
@@ -29,27 +21,28 @@ def all_urls
...
-          end
-        else
-          urls = [@livecheck_args[:url]] if @livecheck_args[:url]
-          urls ||= all_urls
-          regex = @livecheck_args[:regex]
-
-          version_heuristic(livecheckable, urls, regex)
-        end
+      @livecheck_args ||= Livecheck.new
+      version_s = begin
```
samford, line 40: Just so I understand, what are you trying to accomplish with `begin`/`end` here? If you're aiming to restrict the lifetime of `urls`/`regex`, then you should know that `begin`/`end` doesn't affect variable scoping (unfortunately).

If this is the intention, you can simply remove the `begin`/`end` and set `version_s` to the result from `version_heuristic`.

livecheck/extend/formula.rb
```diff
@@ -29,27 +21,28 @@ def all_urls
...
-            result
-          end
-        else
-          urls = [@livecheck_args[:url]] if @livecheck_args[:url]
-          urls ||= all_urls
-          regex = @livecheck_args[:regex]
-
-          version_heuristic(livecheckable, urls, regex)
-        end
+      @livecheck_args ||= Livecheck.new
```
samford, line 39: It seems like this isn't used and could be removed.

nandahkrishna:
> One thing I noticed in testing this is that livecheck blocks don't take precedence over livecheckables here. That is to say, if a formula has a livecheck block and a livecheckable also exists for the formula, livecheck will use the livecheckable.

I realised that this was due to the Livecheckable being loaded and used to `module_eval` from `formulary.rb`, I'm doing some testing after making changes, I'll let you know with the next commit.

cmd/livecheck.rb
```diff
@@ -109,14 +109,12 @@ def print_latest_version(formula)
...
     is_gist = formula.stable.url.include?("gist.github.com")
-    if has_hash_arg_with_skip || formula.livecheck_args == :skip || is_gist
-      skip_msg = if has_hash_arg_with_skip &&
-                    formula.livecheck_args[:skip].is_a?(String) &&
-                    !formula.livecheck_args[:skip].empty?
-        " - #{formula.livecheck_args[:skip]}"
+    if formula.livecheck.skip? || is_gist
+      skip_msg = if formula.livecheck.skip? &&
+                    formula.livecheck.skip_msg.is_a?(String) &&
+                    !formula.livecheck.skip_msg.empty?
```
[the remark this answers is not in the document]
nandahkrishna: Removed the repeated `skip?` and also changed `empty?` to `blank?`.

livecheck/extend/formula.rb
```diff
@@ -29,27 +21,28 @@ def all_urls
...
+      return @livecheck unless block_given? || !arg.empty?
+
+      livecheck_args = block || arg
+      if livecheck_args.is_a? Proc
+        @livecheck.instance_eval(&block)
+      else
+        livecheck_args.each do |key, value|
+          @livecheck.send(key.to_s, value)
+        end
+      end
```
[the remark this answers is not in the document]
nandahkrishna: My current setup for the early return is:
```ruby
return @livecheck unless (block_given? || !arg.empty?) && !livecheckable?
```
And the reason for this is, `formulary.rb` loads the Livecheckable and does a `module_eval`, causing the `@livecheck` instance variable to be overwritten with information from the Livecheckable.

By returning early if `@livecheckable` is true, I prevent the overwrite from happening, but that would mean I can't do the refactoring you said. What do you think of this @samford?

livecheck/extend/formula.rb
nandahkrishna: That was the intention, unfortunately. I'll change it.

livecheck/extend/formula.rb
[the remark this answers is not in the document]
samford: Tinkering with this some more, let's just stick with this general structure (if/else) but make some modifications:

* There isn't any benefit in creating a `livecheck_args` variable here, so we can remove that and simply access the variables directly (`block` and `arg`).
* Update the condition for working with the block to `block_given?`. This _does_ repeat the check (from the early return above) but we'll just have to live with it for the moment. It won't be an issue after we remove hash support in the future.
* Just to keep things more in line with the implementation in Homebrew/brew's `Formula`, move `@livecheckable = true` above the if/else here.

livecheck/extend/formula.rb
```diff
@@ -29,28 +21,24 @@ def all_urls
     end
 
     def livecheck(arg = {}, &block)
-      @livecheck_args = block || arg
+      @livecheck ||= Livecheck.new
+      return @livecheck unless (block_given? || !arg.empty?) && !livecheckable?
```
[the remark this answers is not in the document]
nandahkrishna: Haha yes, got it.

Issue #301, opened 2020-03-04
Add `using` option to explicitly specify the strategy to use


samford:
**NOTE**: There are other feature proposals to come and this primarily applies to niche scenarios, so I wouldn't recommend that anyone take this on just yet. There are more useful features that should be addressed first. I've only created an issue here now because I wanted to avoid bogging down the livecheck DSL issue in Homebrew/brew with too much detail on `using`.

---

The basic idea here is to restructure livecheck's heuristic so it can accept a `:using` option to explicitly specify the strategy to use for a given formula.  The value would be a symbol, corresponding to a strategy in the heuristic (e.g., `:git`, `:sourceforge`, etc.), or an array of these symbols (which livecheck would choose from instead of the pool of all possible strategies).

Though we could have values like `:git` or `:sourceforge`, I imagine this will be most useful for specifying that the default strategy (`page_matches`) should be used instead of a specific strategy. That default strategy would need to be named somehow like `:default` or `:page`.  Some people may erroneously think that `:default` means "the strategy that livecheck would normally choose for this URL", so something like `:page` may be better.

`using` should work both with or without a URL.  When a URL is provided, livecheck would use the specified strategy without going through the built-in strategy-guessing (based on the URL).  When a URL isn't provided, livecheck would gather the URLs from the formula (like it already does) but only use the specified strategy (or strategies) when checking each URL.

It looks like this in practice:

```ruby
class SomeSoftware
  livecheck :url   => "https://www.example.com/downloads",
            :using => :page,
            :regex => /somesoftware-(\d+(?:\.\d+)+)\.tar/
end
```

Assuming we had a specific strategy for example.com URLs, this would override it and use `page_matches` instead with the regex.

The two main applications I imagine for `using` are detailed below.

## Overriding `livecheck`'s Built-In Strategy Selection

`livecheck` does two types of guessing when looking for versions:

* If a URL isn't provided in a livecheckable, it goes through the URLs in the formula (in order: HEAD, stable, stable mirrors, homepage) until it gets a result.
* With each URL, it decides what strategy to apply to the URL by going through a big `if` statement.

We can already tell livecheck to only check a specific URL but it decides for itself what strategy to use.

[...]

PR #569, opened 2020-05-02
aspell: add livecheckable

miccal:
My first contribution, be kind if I have made a mistake :)

### DIFF
--- a/Livecheckables/aspell.rb
+++ b/Livecheckables/aspell.rb
@@ -1,0 +1,4 @@
+class Aspell
+  livecheck :url   => "https://ftp.gnu.org/gnu/aspell/",
+            :regex => /href="aspell-([0-9\.]+)\.t/
+end
--- a/Livecheckables/aspell.rb
+++ b/Livecheckables/aspell.rb
@@ -1,4 +1,4 @@
 class Aspell
   livecheck :url   => "https://ftp.gnu.org/gnu/aspell/?C=M&O=D",
-            :regex => /href="aspell-([0-9\.]+)\.t/
+            :regex => /href=.+aspell-v?(\d+(?:\.\d+)+)\.t/
 end
--- a/Livecheckables/aspell.rb
+++ b/Livecheckables/aspell.rb
@@ -1,4 +1,4 @@
 class Aspell
-  livecheck :url   => "https://ftp.gnu.org/gnu/aspell/",
+  livecheck :url   => "https://ftp.gnu.org/gnu/aspell/?C=M&O=D",
             :regex => /href="aspell-([0-9\.]+)\.t/
 end


Livecheckables/aspell.rb
```diff
@@ -0,0 +1,4 @@
+class Aspell
+  livecheck :url   => "https://ftp.gnu.org/gnu/aspell/",
```
samford, line 2:
```suggestion
  livecheck :url   => "https://ftp.gnu.org/gnu/aspell/?C=M&O=D",
```

With index pages, we can sometimes append the `?C=M&O=D` query string parameters to sort the items on the page by their last modified date in descending order. This doesn't impact livecheck's ability to identify the newest version but it's nice to have them in order if we ever check the debugging output (e.g., `brew livecheck --debug aspell`).

Livecheckables/aspell.rb
```diff
@@ -0,0 +1,4 @@
...
+            :regex => /href="aspell-([0-9\.]+)\.t/
```
samford, line 3:
```suggestion
            :regex => /href=.+aspell-v?(\d+(?:\.\d+)+)\.t/
```

One thing to note here is that we try to avoid using the old `[0-9\.]+` regex to match versions. You'll find this in some old livecheckables but one day we'll go back and replace them all with the prevailing regex we use nowadays.

The reason why we avoid it is because it's too loose and doesn't accurately represent our intention. This regex can match weird things like `......`, `0000000`, etc. when we our intention is to match versions like `1.2`, `1.2.3`, etc.

The snippet that we use for this nowadays is `v?(\d+(?:\.\d+)+)`. This allows for an optional `v` before the numeric version and only matches versions like `1.2` or longer (e.g., `1.2.3`, `1.2.3.4`).

Even when the strings/tags don't include a leading `v`, we still include the full regex snippet with the optional `v?` in case upstream either starts adding the leading `v` or accidentally pushes a release with a leading `v`. Basically, we try to make checks less brittle and more futureproof when we can.

You'll see this snippet as part of regexes (as in this PR) or as the entire regex (e.g., `^v?(\d+(?:\.\d+)+)$` for Git tags) pretty often in PRs these days, as I've pushed it as a better standard than `[0-9\.]+`. It's a definitely good idea to become friends with the "standard" regex as I'm likely to recommend it when reviewing PRs.

Besides that, when we're matching within `href` attributes, I usually start with `href=.+`.

One reason is that some websites will use single quotes (`'`) around the attribute instead of double quotes (`"`); you're supposed to use double quotes but every once in a blue moon we see single quotes. Using `.+` instead of `"` makes it match both `"` and `'`.

The other reason is that `.+` will also match anything preceding the file name in the URL. Index pages like this usually use relative URLs  with nothing preceding the filename but if we relied on that and something changed upstream, livecheck wouldn't be able to match versions until we updated the regex. This also helps to make the check a little less brittle.

miccal:
Thank you @samford for taking the time to explain -- I just copied what I had found in similar files that used `ftp.gnu.org`.

samford:
> Thank you @samford for taking the time to explain -- I just copied what I had found in similar files that used `ftp.gnu.org`.

No worries, I started working on livecheck in the same way and it was only through working on the project and building experience that I came up with some better standards. Unfortunately not all of the existing livecheckables are up to the current standards but it's a constant work in progress. Once I implement some forthcoming features, I'll hopefully be able to get back to improving the existing livecheckables (and fixing checks that don't automatically work via the heuristic).

I usually recommend that new contributors look through [livecheck/heuristic.rb](https://github.com/Homebrew/homebrew-livecheck/blob/master/livecheck/heuristic.rb), to understand livecheck's built-in checks. If you see "(guesed)" in the output for a check, it means that the check is simply using livecheck's heuristic and that a livecheckable isn't involved. In some cases, we let the heuristic take care of most of the work and simply add a regex to adjust the matching, so there's value in knowing how it works.

miccal:
> If you see "(guesed)" in the output for a check, it means that the check is simply using livecheck's heuristic and that a livecheckable isn't involved. 

@samford,so just to be clear, if `brew livecheck` outputs a result with `(guessed)`, you would prefer them NOT to have a `Livecheckables` file, unless the "guess" is way off?

samford:
> ...if `brew livecheck` outputs a result with `(guessed)`, you would prefer them NOT to have a `Livecheckables` file, unless the "guess" is way off?

Yep, the heuristic is capable of automatically handling checks for a good number of formulae and we only add a livecheckable if we encounter something that needs to be addressed (e.g., an incorrect version or weird result, error, etc.). Changes to the heuristic can improve the situation for a number of formulae (as in the recent SourceForge PR) but only if they don't fully override the heuristic's behavior (i.e., providing both a URL and regex in a livecheckable). Otherwise, we have to make modifications to livecheckables one by one, so we try to lean on the heuristic as much as possible.

As a practical example, one of the strategies that the heuristic can use checks the project's SourceForge RSS feed for versions. However, other software in separate folders is also included in the RSS feed, so the default regex for the strategy may match versions for software other than what we intend and cause livecheck to give an incorrect version as newest. In this case, the URL and strategy the heuristic uses is good but we would need to add a livecheckable with a more specific regex to restrict matching to only the versions we want. The heuristic still uses the SourceForge strategy but it instead uses the regex supplied in the livecheckable.

---

It may help to mention that livecheck only recently (as of fall 2019) started getting more significant attention again. Things have improved since then but there's still around 950 formulae that give an error, so there's quite a ways to go before livecheck's in good shape.

However, we're not going to do a big push to tackle those until after I've finished implementing livecheckable-related features in the coming weeks (most importantly, the `using`/`not_using` feature in #301). The forthcoming features impact how we write livecheckables, so fixing the checks for those formulae now would give us 950 more livecheckables that we would need to go through and potentially modify after features are implemented.

It's fine to contribute in the interim time if there are checks you want to fix for particular formulae but I only mention this to discourage _large_ numbers of contributions for the immediate future. It's easy to get bitten by livecheck and the next thing you know there are 10-20 PRs open, so we're just trying to avoid that extreme for a brief moment. If you don't intend to create a lot of PRs, then just don't worry about it at all :)

Sorry for the long answer but hopefully this is helpful information.

PR #587, opened 2020-05-03
pass: add livecheckable

miccal:
Before:
```
-bash-5.0.17- /Users/miccal (32) [> brew livecheck pass
Error: pass: Unable to get versions
```
After:
```
-bash-5.0.17- /Users/miccal (32) [> brew livecheck pass
pass : 1.7.3 ==> 1.7.3
```

### DIFF
--- a/Livecheckables/pass.rb
+++ b/Livecheckables/pass.rb
@@ -1,0 +1,4 @@
+class Pass
+  livecheck :url   => "https://git.zx2c4.com/password-store/refs/",
+            :regex => /href=.+password-store-v?(\d+(?:\.\d+)+)\.t/
+end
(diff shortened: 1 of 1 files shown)


samford:
One bit of feedback about the original regex: I recently discovered in testing that `href=.+something` can potentially escape the bounds of the `href` attribute. `href=.+` was originally a replacement for `href=".*?` (which was non-greedy to avoid this particular situation) but I didn't think that it would be necessary to restrict `.+` to be non-greedy. Unfortunately, it _is_ necessary and I'm going to have to go back and edit a bunch of livecheckables with this in mind 😭

This is new information, so I just wanted to give you a heads up to be sure to use `href=.+?` in the future.

miccal:
Thanks again for your help @samford.

samford:
> Thanks again for your help @samford.

You're welcome. Thanks for the contribution!

PR #594, opened 2020-05-05
vim: add livecheckable

miccal:
As stated in the [Formula for vim](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vim.rb):
```
# vim should only be updated every 50 releases on multiples of 50
```

This `Livecheckables` file will only output those `version`'s whose last two digits are either `00` or `50`, which I confirmed using the `--debug` flag.

Output before:
```
-bash-5.0.17- /Users/miccal (32) [> brew livecheck vim
vim (guessed) : 8.2.0654 ==> 8.2.0695
```

Output after:
```
-bash-5.0.17- /Users/miccal (32) [> brew livecheck vim
vim : 8.2.0654 ==> 8.2.0650
```

Note that currently the Formula for `vim` is at `8.2.0654` because of a [problem with `8.2.0650`](https://github.com/Homebrew/homebrew-core/pull/53849), and I quote:

> Build is broken in 8.2.0650, fixed in 8.2.0654. Next bump should resume normal schedule with 8.2.0700.

### DIFF
--- a/Livecheckables/vim.rb
+++ b/Livecheckables/vim.rb
@@ -1,0 +1,4 @@
+class Vim
+  livecheck :url   => "https://github.com/vim/vim.git",
+            :regex => /^v?(\d+(?:\.\d+)+\d{2}(?<=00|50))$/
+end


samford:
I appreciate the effort but the goal of livecheck is to identify the latest version of a formula's software, not the latest appropriate version with respect to Homebrew's update throttling. I'm against the idea of artificially restricting livecheck from displaying the latest version, so I won't be merging this, sorry.

I'll take the time to review the code before closing this (for your benefit), as there's one thing that I would have changed.

Livecheckables/vim.rb
```diff
@@ -0,0 +1,4 @@
+class Vim
+  livecheck :url   => "https://github.com/vim/vim.git",
+            :regex => /^v?(\d+(?:\.\d+)+\d{2}(?<=00|50))$/
```
samford, line 3:
```suggestion
            :regex => /^v?(\d+(?:\.\d+)+(?:00|50))$/
```

In my testing, `(?:00|50)` matched the same versions without being as verbose or requiring a lookbehind, so this could have been simplified if the PR was appropriate for livecheck.

miccal: Thanks for the feedback @samford -- I was not sure if the `$` (which as you said would ensure that it did not match a tag like `1.2.3-rc1`) would work all the time with just a `(?:00|50)`, hence why I used a lookbehind with `\d{2}`to ensure that it performed the lookbehind after digits only. More to learn :)

miccal:
@samford no problem at all with not merging this -- I thought this may be helpful as I read that a long-term goal was to [automatically create pull requests based on `livecheck`](https://github.com/Homebrew/gsoc#automatically-create-pull-requests-based-on-homebrewlivecheck).

Anyway, it was fun experimenting with this and learning about lookarounds.

Livecheckables/vim.rb
samford: `$` basically means "the end of the line", which is also the end of the string in the context of Git tags. `(?:00|50))$` ensures that the last part of the string is `00` or `50` and since there's nothing between `\d+(?:\.\d+)+` and `(?:00|50)`, it would always be the last digits of the last part of the version (e.g., 1.2.300, 1.2.350, 1.2.3.000, etc.). With that in mind, it wouldn't match something like `1.2.300-rc1`.

The only thing to note is that this would fail to match a version with fewer than three digits in the last part of the version, like `1.2.00` or `1.2.50`. The Git tags use a minimum of three digits in the last part of the version and seem to have standardized on four nowadays, so it wouldn't be a practical concern yet. However, if we wanted to address that potential issue for the sake of future reliability, we could be more explicit about the final part of the version: `^v?(\d+(?:\.\d+)+\.\d*(?:00|50))$`

miccal: Thank you again for the explanation, your efforts are very much appreciated.

samford:
For what it's worth, it's best to avoid lookarounds if you can, since they can be more expensive from a computational perspective. That is to say, if you can accomplish the same thing without a lookaround, it's preferable. Every now and again they're required to properly match something but there aren't any in the existing livecheckables and I haven't yet found a case where one was absolutely necessary.

While I have your ear, I noticed that you tend to create duplicate PRs, seemingly closing your open PR to make a new PR with changes. You can modify your existing PR by editing the PR title or text, as well as pushing (or force pushing) changes to your branch, so there's generally no need to close a PR to open a new one. Only creating one PR helps to keep things tidy on our end. Thanks.

PR #601, opened 2020-05-07
Move site-specific logic to separate files

maxim-belkin:
First step toward reorganizing and improving homebrew/livecheck.
--EDIT--
I kept changes to a minimum so that it's easy to see what was done.


samford:
I've been planning to do something similar when I implement the `using`/`not_using` feature (after I finally finish moving the open SourceForge PR to completion in the coming days). I'm going to experiment with different ways of handling this particular aspect of the `using` feature, so I can't yet say whether this is the best way of approaching this.

We can either try to use your PR as a first step towards that feature (keeping the PR open and revisiting it once I'm done experimenting) or we can simply close this in favor of handling it as part of the forthcoming `using` feature. What do you think?

maxim-belkin:
This PR, the sourceforce PR, and the `using` feature proposal are orthogonal to each other: you can implement one without implementing the other two. Currently, sourceforce PR is still in progress and there is no timeline for the `using` feature. Because you were planning to do something similar, I would suggest to:

1. merge this PR.
2. rebase & (once it's ready) merge sourceforce PR 
3. implement the `using` feature.

samford:
I meant "something similar" in a broad sense, not "exactly this implementation". This PR isn't ready to merge right now and needs changes, which I'll explain when I review this in the future. Besides that, I would like to research this a bit and explore alternatives, so I can properly assess whether this is exactly how we should proceed.

The only reason I mentioned the SourceForge PR is that I won't be able to get to this until after I'm done with it. The reason to merge the SourceForge PR first isn't technical and is rather because it has been open for several weeks and I've been actively working to move it forward over the past week or so. It's basically done right now, I just have to create and merge a bunch of livecheckable PRs (already implemented locally), rebase the SourceForge PR, and merge it.

My primary purpose in commenting here was to let you know that I've seen this (instead of leaving you in limbo) and to understand whether it was worth keeping the PR open. You understandably want this to move forward, so I'll revisit it when I can (probably less than a week from now but don't hold me to it).

maxim-belkin:
@samford, I've updated the PR.

1. Major change: use a single `case` structure to determine which method to use and call it with `.send`.
2. Replaced 
    ```ruby
    page_matches(url, regex).each do |match|
      version = Version.new(match)
      match_version_map[match] = version
    end
    ```
    with a one-liner `page_matches(url, regex).map { |match| [match, Version.new(match)] }.to_h`
3. Replaced `git_tags_only_debian` method with a one-line check
    ```ruby
    tags.all? { |tag| tag.startwith?("debian/") }
    ```

As before, this PR moves site-specific methods to separate files in ` livecheck/heuristics`, e.g. `livecheck/heuristics/bitbucket.rb`, livecheck/heuristics/python.rb`, etc.

samford:
I recently finished the implementation of the `using` feature (which we'll probably refer to as `strategy` in livecheckables), so I'm in a position to better review this now. While there are _some_ similarities between this PR and my local changes, I went about this in a notably different manner.

It would take time/effort to rework this into something that I can reasonably rebase onto but I'm not sure that it would be a worthwhile investment of time, since my work with the `using` feature meets the same goals (and in a manner that should integrate more easily with Homebrew/brew in the near future). That said, it's _technically_ possible and I don't want you to feel like you wasted your time here, so I'll give it a try.

Would you mind if I push commits to this PR's branch to rework this into something that more closely resembles the changes in the forthcoming `using` feature? I'll explain everything afterward but it will be faster for me to just rework this myself (while referencing my local changes). I would review this properly if I had more time (going through it and leaving comments/suggestions) but I'm trying to make this work with the time I have available.

samford:
I haven't heard back about this but I'm going to merge my reworked version of this PR, as I need to wrap this up to continue moving forward.

In general, I've tried to reduce the changes here to simple restructuring and left out the parts that rework strategies. This makes it easy to reason about the changes here and we can always address modifications to strategies in future PRs.


### Notable Changes

The main thing here is that I've put all the strategies in one `livecheck_strategy.rb` file and appropriately named the functions `*_strategy`, as these are strategies rather than "heuristics".  This is a little closer to what I'm doing with the `strategy` feature (but it takes this even further), so this will help with rebasing.

The reason why one file is potentially better is because livecheck will be integrated into Homebrew/brew in the not-too-distant future, so we have to be aware of how it will fit in that context. What I'm doing in the `strategy` feature is somewhat similar to the existing `download_strategy.rb` (where all the strategies are classes and appear in one file), so it should make sense in Homebrew/brew.

* I reverted the changes to the strategies themselves but I specifically removed the `page_matches` one-liner because it tripled related memory allocations compared to the existing approach. It's good to profile changes like these to make sure they don't negatively impact resource usage or performance before proposing them in a PR.

* I reinstated the current `if`/`else` conditions instead of the `case` statement. You moved some of these conditions into the strategies as early returns (e.g., SourceForge special cases) but this isn't something we want to do because it prevents the heuristic from working properly. Currently, the heuristic will fall through to a different strategy if conditions aren't met (defaulting to the `page_match` strategy if all else fails). With your changes, the strategy is executed and simply returns nothing, so the heuristic can't try another strategy. That said, I do use a `case` statement in the `strategy` feature for the heuristic's strategy-selection code. The way I handle the conditions doesn't lead to this problem, so it's best if we just sidestep the issue entirely in this PR.

* I renamed `github_preprocess()` to a more general name, `preprocess_url`, as it deals with more than just GitHub-related changes (e.g., the `/-/archive/` part is for Bitbucket URLs). This function is where we would make any changes in the URL before working with it, so it makes sense to use a general name. I placed this in `heuristic.rb` for now, since it doesn't make sense in `livecheck_strategy.rb`. It may be moved elsewhere in the long run but this will be fine for now.

* I've moved some of the frozen arrays out of functions/loops and made them constants, so they're only allocated once. I'll be moving any remaining ones (e.g., the prerelease version strings) into constants in follow-up PRs. [The `strategy` PR will improve this situation even further, for what it's worth.]

* Added a `next` statement for when no strategy is appropriate for a URL and no `regex` is provided, to prevent `Symbol.send` from executing.


### Next Steps

Outside of what's mentioned above, I believe the only change here that I would like to see a separate PR for is the refactoring/inlining of `git_tags_only_debian?` into the `git` strategy (since it's short and only used there). The one thing to note is that your code seemingly had a typo (`startwith?` instead of `start_with?`), so that should be addressed in the process.

I hope this ends up being an amenable solution to this PR. It would have been much easier to simply close this in favor of the forthcoming `strategy` PR but I didn't want you to feel like you wasted your time or I didn't value your work here.

That said, I can ask one of the Slack admins to add you to the `#livecheck` channel if you're interested, so we can try to discuss things beforehand in hopes of avoiding duplicated effort in the future.

PR #744, opened 2020-05-19
consul: remove url and update regex

dawidd6:
previously: https://releases.hashicorp.com/consul/1.8.0-beta1
now: https://releases.hashicorp.com/consul/1.7.3/consul_1.7.3_SHA256SUMS


samford:
Thanks, Dawid!

PR #747, opened 2020-05-22
archey: update url

nandahkrishna:
The `url` in `archey.rb`, earlier `https://github.com/obihann/archey-osx.git`, caused a style issue when migrated to a `livecheck` block in `homebrew-core`, as it seemed like a binary package rather than a source package (although it is a source package).

The `url` in the Livecheckable has been updated to `https://github.com/obihann/archey-osx/releases/`, which works with `livecheck` and does not cause a style issue when migrated to `homebrew-core`.


nandahkrishna:
Oops, it does cause a style issue in `homebrew-core`. I seem to have missed it when checking the output of `brew style homebrew/core`. My bad.

Closing this in favour of a change in Homebrew/brew instead.

PR #748, opened 2020-05-22
freetds: update url

nandahkrishna:
The `url` in `freetds.rb`, earlier a `fossies.org` url, caused a style issue when migrated to a `livecheck` block in `homebrew-core`.

The `url` in the Livecheckable has been updated to http://www.freetds.org/files/stable/, which works with `livecheck` and does not cause a style issue when migrated to `homebrew-core`.

### DIFF
--- a/Livecheckables/freetds.rb
+++ b/Livecheckables/freetds.rb
@@ -1,4 +1,4 @@
 class Freetds
-  livecheck :url   => "https://fossies.org/linux/privat/",
-            :regex => /HREF="freetds-([\d.]+)\.tar/
+  livecheck :url   => "http://www.freetds.org/files/stable/",
+            :regex => /href=.*?freetds-v?(\d+(?:\.\d+)+)\.t/i
 end


Livecheckables/freetds.rb
```diff
@@ -1,4 +1,4 @@
 class Freetds
-  livecheck :url   => "https://fossies.org/linux/privat/",
-            :regex => /HREF="freetds-([\d.]+)\.tar/
+  livecheck :url   => "http://www.freetds.org/files/stable/",
+            :regex => /freetds-([\d.]+)\.tar\.gz/
```
samford, line 3:
```suggestion
            :regex => /href=.+?freetds-v?(\d+(?:\.\d+)+)\.t/i
```

There are a few things that can be improved here, while we're editing this livecheckable:

* We almost always want to search within the `href` attribute on archive index pages like these, so removing that from the regex here isn't appropriate. We can just make it lowercase and/or make the entire regex case insensitive.
* We can update the part that matches the numeric version `([\d.]+)` to the more explicit regex that we use for this (`v?(\d+(?:\.\d+)+)`), which only matches what we want (and wouldn't match something weird like `.....`).
* There's nothing to say that the archive format couldn't become `tgz` or `tar.xz` in the future, which wouldn't work with either the old or the new regex in this PR. We just use `\.t` to match extensions like `.tgz`, `.tar.*`, etc, so there's less of a chance of it breaking in the future.

PR #767, opened 2020-05-27
spice-protocol: update regex

nandahkrishna:
The earlier regex was too specific w.r.t the extension and was not matching the latest version (as they had updated their compression format). Updated the regex to be more loose w.r.t the extension, while also updating the version number matching to our preferred regex.

### DIFF
--- a/Livecheckables/spice-protocol.rb
+++ b/Livecheckables/spice-protocol.rb
@@ -1,6 +1,6 @@
 class SpiceProtocol
   livecheck do
     url "https://www.spice-space.org/download/releases/"
-    regex(/spice-protocol-([\d.]+)\.tar\.bz2/)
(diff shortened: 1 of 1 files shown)


Livecheckables/spice-protocol.rb
```diff
@@ -1,6 +1,6 @@
 class SpiceProtocol
   livecheck do
     url "https://www.spice-space.org/download/releases/"
-    regex(/spice-protocol-([\d.]+)\.tar\.bz2/)
+    regex(/spice-protocol-v?(\d+(?:\.\d+)+)\.t/)
```
samford, line 4:
```suggestion
    regex(/href=.*?spice-protocol-v?(\d+(?:\.\d+)+)\.t/)
```

We should update this to only match within `href` attributes while we're at it.

nandahkrishna:
Modified the regex to match only within the href attributes.

PR #768, opened 2020-05-27
plzip: update url and regex

nandahkrishna:
Fixes a `redirection-forbidden` error for `plzip`.


nandahkrishna:
Modified the regex.

PR #770, opened 2020-05-27
Remove commands/cat.rb

samford:
I'm not sure if there's value in keeping this around and removing it means there's one less thing to migrate to Homebrew/brew in the future.

Are there any objections to this?


nandahkrishna:
I think we can remove this. There's already the `brew cat` developer command that outputs the source code of a Formula from homebrew-core. Now that all `livecheck` blocks will be migrated to Formulae, this is simply not required and already taken care of in Homebrew/brew, if I'm not wrong.

samford:
Good points. Thanks for the input, Nanda.

I'm going to go ahead and merge this, with that in mind.

PR #788, opened 2020-06-01
gradle: add livecheckable

nandahkrishna:
Add a Livecheckable for `gradle`, using their GitHub repository `url` to check for newer versions.

### DIFF
--- a/Livecheckables/gradle.rb
+++ b/Livecheckables/gradle.rb
@@ -1,0 +1,6 @@
+class Gradle
+  livecheck do
+    url "https://services.gradle.org/distributions/"
+    regex(/href=.*?gradle-v?(\d+(?:\.\d+)+)-all\.(?:[tz])/i)
+  end
+end


Livecheckables/gradle.rb
```diff
@@ -0,0 +1,6 @@
+class Gradle
+  livecheck do
+    url "https://github.com/gradle/gradle.git"
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url "https://services.gradle.org/distributions/"
    regex(/href=.*?gradle-v?(\d+(?:\.\d+)+)-all\.(?:[tz])/i)
```

The formula uses an archive file from https://services.gradle.org/distributions/, so it's more appropriate to check that index page.

Though the Git tags are convenient, we try to check as close to the formula's stable archive location as possible. If that's not possible or there's an issue with that approach, then we start checking other sources (e.g., Git tags).

nandahkrishna:
Made the changes suggested.

PR #789, opened 2020-06-01
telnet: add livecheckable

nandahkrishna:
Adding a Livecheckable for `telnet`. Important note: homebrew-core lists the version for `telnet` as that of Apple's `remote_cmd` version (currently 63). This Livecheckable considers the version of `remote_cmd` to be `telnet`'s version as well.

### DIFF
--- a/Livecheckables/telnet.rb
+++ b/Livecheckables/telnet.rb
@@ -1,0 +1,6 @@
+class Telnet
+  livecheck do
+    url "https://opensource.apple.com/tarballs/remote_cmds/"
+    regex(/href=.*?remote_cmds-v?(\d+(?:\.\d+)*)\.t/i)
+  end
+end


Livecheckables/telnet.rb
```diff
@@ -0,0 +1,6 @@
+class Telnet
+  livecheck do
+    url "https://opensource.apple.com/source/remote_cmds/"
+    regex(/href=.*?remote_cmds-v?(\d+(?:\.\d+)*)/i)
```
samford, line 4:
```suggestion
    url "https://opensource.apple.com/tarballs/remote_cmds/"
    regex(/href=.*?remote_cmds-v?(\d+(?:\.\d+)*)\.t/i)
```

The formula uses an archive from https://opensource.apple.com/tarballs/remote_cmds/, so it seems appropriate to check that index page for new versions. Was there a reason to check the "source" directories instead?

PR #791, opened 2020-06-02
fix style

dawidd6:
https://github.com/Homebrew/brew/pull/7677


samford:
The CI livecheck failures are just SSL failures, rate limit errors, and otherwise known issues with existing livecheckables.

Thanks for this, Dawid! This was one of the items on my clean-up to-do list, so it's kind of nice that we have an audit for it now.

PR #792, opened 2020-06-02
mysql-client: add livecheckable

nandahkrishna:
Fixes `livecheck` error by adding a Livecheckable for `mysql-client`; using GitHub repo to match versions tagged as `mysql-x.x.x` and not the MySQL Cluster versions. Concerns: There's a version `mysql-client@5.7` formula as well. Do let me know if changes are required.

### DIFF
--- a/Livecheckables/mysql-client.rb
+++ b/Livecheckables/mysql-client.rb
@@ -1,0 +1,6 @@
+class MysqlClient
+  livecheck do
+    url "https://github.com/mysql/mysql-server.git"
+    regex(/^mysql-v?(\d+(?:\.\d+)+)$/i)
+  end
+end


Livecheckables/mysql-client.rb
```diff
@@ -0,0 +1,6 @@
+class MysqlClient
+  livecheck do
+    url "https://github.com/mysql/mysql-server.git"
+    regex(/mysql-v?(\d+(?:\.\d+)+)/i)
```
samford, line 4:
```suggestion
    regex(/^mysql-v?(\d+(?:\.\d+)+)$/i)
```

Adding the `$` delimiter at the end of the regex for Git tags makes sure we don't match any potential prerelease versions in the future, if they ever appear (e.g., `1.2.3-rc1`, `1.2.3-alpha1`, `1.2.3-beta1`, etc.). We also use the leading `^` delimiter, to make sure we don't end up matching any weird tags that could pop up in the future (e.g., `weird-mysql-1.2.3`).

PR #794, opened 2020-06-02
scala: add livecheckable

nandahkrishna:
Adding a Livecheckable for `scala` using the `url` https://www.scala-lang.org/files/archive.

### DIFF
--- a/Livecheckables/scala.rb
+++ b/Livecheckables/scala.rb
@@ -1,0 +1,6 @@
+class Scala
+  livecheck do
+    url "https://www.scala-lang.org/files/archive"
+    regex(/href=.*?scala-v?(\d+(?:\.\d+)+)(?:.final)?\.t/i)
+  end
+end


Livecheckables/scala.rb
```diff
@@ -0,0 +1,6 @@
+class Scala
+  livecheck do
+    url "https://www.scala-lang.org/files/archive"
+    regex(/href=.*?scala-v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?scala-v?(\d+(?:\.\d+)+)(?:.final)?\.t/i)
```

Some of the older archives have names like `scala-2.5.0-final.tar.gz` or `scala-2.8.2.final.tgz` (note that one style uses `-final` and the other uses `.final`, so I used `.final` in the regex to match either). It wouldn't hurt to update the regex here to optionally match this file name style, in case they ever suddenly use it again in the future.

[I initially thought this PR wouldn't need changes, so I had rebased it in anticipation of merging it. I saved you the work for this PR, at least, ahah.]

nandahkrishna: Haha, thanks 😅

PR #796, opened 2020-06-02
metis: add livecheckable

nandahkrishna:
Using the `homepage` as `url` and matching text. There doesn't seem to be a proper index page of *all* versions, although a tree of some versions does exist.

### DIFF
--- a/Livecheckables/metis.rb
+++ b/Livecheckables/metis.rb
@@ -1,0 +1,6 @@
+class Metis
+  livecheck do
+    url "http://glaros.dtc.umn.edu/gkhome/metis/metis/download"
+    regex(%r{href=.*?/metis.v?(\d+(?:\.\d+)+)\.t}i)
+  end
+end


Livecheckables/metis.rb
```diff
@@ -0,0 +1,6 @@
+class Metis
+  livecheck do
+    url :homepage
+    regex(/METIS\ stable\ version:\ v?(\d+(?:\.\d+)+)/i)
```
samford, line 4:
```suggestion
    url "http://glaros.dtc.umn.edu/gkhome/metis/metis/download"
    regex(%r{href=.*?/metis.v?(\d+(?:\.\d+)+)\.t}i)
```

It's probably better to use the download page here, since it contains the link to the stable archive file used in the formula. [I had to include the `/` before the filename here to prevent the regex from matching the `mt-metis` release as well.]

nandahkrishna: Ah okay, thanks!

PR #797, opened 2020-06-03
xmlto: add livecheckable

miccal:
Output before:
```
-bash-5.0.17- /Users/miccal (28) [> brew livecheck xmlto
Error: xmlto: Unable to get versions
```
Output after:
```
-bash-5.0.17- /Users/miccal (28) [> brew livecheck xmlto
xmlto : 0.0.28 ==> 0.0.28
```

I hope I have done this right -- the syntax has changed since I last contributed.

Thanks.

### DIFF
--- a/Livecheckables/xmlto.rb
+++ b/Livecheckables/xmlto.rb
@@ -1,6 +1,6 @@
 class Xmlto
   livecheck do
     url "https://releases.pagure.org/xmlto/?C=M&O=D"
-    regex(/href=.+xmlto-v?(\d+(?:\.\d+)+)\.t/)
+    regex(/href=.*?xmlto-v?(\d+(?:\.\d+)+)\.t/i)
   end
 end
--- a/Livecheckables/xmlto.rb
+++ b/Livecheckables/xmlto.rb
@@ -1,0 +1,6 @@
+class Xmlto
+  livecheck do
+    url "https://releases.pagure.org/xmlto/?C=M&O=D"
+    regex(/href=.+xmlto-v?(\d+(?:\.\d+)+)\.t/)
+  end
+end


Livecheckables/xmlto.rb
```diff
@@ -0,0 +1,6 @@
+class Xmlto
+  livecheck do
+    url "https://releases.pagure.org/xmlto/?C=M&O=D"
+    regex(/href=.+xmlto-v?(\d+(?:\.\d+)+)\.t/)
```
samford, line 4:
```suggestion
    regex(/href=.*?xmlto-v?(\d+(?:\.\d+)+)\.t/i)
```

There are a couple updates in our regex standards here to make you aware of:

I've settled on using `href=.*?` for the moment (instead of `href=.+?`), since I once encountered a page that didn't use any quotation marks around attributes (the horror!). That said, regardless of whether we use `.*` or `.+`, we want to make sure to make it a non-greedy match with `?` like `href=.*?`. If we don't, the greedy match can easily escape the bounds of the `href` attribute and cause trouble for us.

I've also decided to use case insensitive regexes unless we specifically need case sensitivity for a specific check. This is the `i` at the end of the regex literal here (i.e., after the closing `/` or `}`). Basically, this will help to reduce the chances of a check breaking simply due to a change in text case.

miccal: Thank you (as always) for the help and explanation @samford.

samford:
Thanks for your work here, @miccal! It's good to see you didn't have any trouble with the new block syntax.

Just to give you a heads up, livecheckables will be incorporated into their respective formulae in Homebrew/homebrew-core in the not-too-distance future (sometime in the next few months). Updating the existing livecheckables to use the new block format was one step in this process.

We have an ongoing Google Summer of Code project where we have a student, Nanda, working on certain parts of livecheck. He was responsible for implementing the `livecheck` DSL that we're using now and he updated the existing livecheckables to the new format using a script he wrote.

He has another script to migrate the livecheckables to homebrew-core formulae but he's waiting for me to finish bringing the existing livecheckables up to our current standards. Once that's done, we'll do the actual migration.

So keep an eye out for that. If you come to this repo in the future and the livecheckables are missing or the repo's archived, you'll know what's happened.

PR #799, opened 2020-06-03
popt: add livecheckable

nandahkrishna:
The rpm5.org url is down (has been for a while probably, as apparent from the use of a Wayback Machine url in homebrew-core). They do have rpm.org now, I think, and this GitHub repository for `popt`. I tried not using a regex and it works too, so let me know if I should add one.

### DIFF
--- a/Livecheckables/popt.rb
+++ b/Livecheckables/popt.rb
@@ -1,0 +1,11 @@
+class Popt
+  # Until rpm.org releases a new version beyond 1.16, it's unclear whether
+  # checking the rpm.org archive index (http://ftp.rpm.org/mirror/popt/) would
+  # give us the latest version. As of writing this, there's a 1.18 release
+  # candidate at: http://ftp.rpm.org/releases/testing/popt-1.18-rc1.tar.gz
+  # Checking the Git repo seems like a safe bet until we see how this plays out.
+  livecheck do
+    url "https://github.com/rpm-software-management/popt.git"
+    regex(/^(?:popt)?-v?(\d+(?:[._-]\d+)+)(?:-release)?$/i)
+  end
+end


Livecheckables/popt.rb
```diff
@@ -0,0 +1,5 @@
+class Popt
+  livecheck do
+    url "https://github.com/rpm-software-management/popt.git"
+  end
```
samford, line 4:
```suggestion
  # Until rpm.org releases a new version beyond 1.16, it's unclear whether
  # checking the rpm.org archive index (http://ftp.rpm.org/mirror/popt/) would
  # give us the latest version. As of writing this, there's a 1.18 release
  # candidate at: http://ftp.rpm.org/releases/testing/popt-1.18-rc1.tar.gz
  # Checking the Git repo seems like a safe bet until we see how this plays out.
  livecheck do
    url "https://github.com/rpm-software-management/popt.git"
    regex(/^(?:popt)?-v?(\d+(?:[._-]\d+)+)(?:-release)?$/i)
  end
```

Per our discussion on Slack, we can update the formula to use the GitHub repo as the homepage and a stable archive from rpm.org (http://ftp.rpm.org/mirror/popt/popt-1.16.tar.gz). I'll open a PR in homebrew-brew for these changes shortly.

It's not yet clear what will happen once the next version beyond 1.16 is released (e.g., will it be in /mirror/popt/ or somewhere in /releases/), so it's probably safest to check the GitHub repo for now.

With this in mind, we should add a comment to explain the situation and also use a regex to restrict matching a bit (potentially avoiding unstable releases).

nandahkrishna: Alright 👍🏼

miccal:
Hello @samford and @nandahkrishna, I have noticed for `popt` that `livecheck` reports:
```
-bash-5.0.17- /Users/miccal (30) [> brew livecheck popt
popt : 1.16 ==> 1_16
```
Note the `.` and `_` in the two reported versions -- I am not sure if this was intended?

Thank you.

nandahkrishna:
It's not yet possible to change the `_` to a `.`, and the reason you see this output is because the Formula is versioned that way (the text we match, to be precise, uses `_` instead of `.`).

miccal:
Thank you for the explanation @nandahkrishna.

PR #800, opened 2020-06-03
libspatialite: add livecheckable

nandahkrishna:
They happen to have just one version (the latest) with an `a` affixed, so instead of using `\d` I used `[0-9a-z]` for the patch part alone, but it could possibly be improved.


nandahkrishna:
Updated 👍🏼

PR #801, opened 2020-06-03
openssh: add livecheckable

nandahkrishna:
Adding Livecheckable for `openssh`.

### DIFF
--- a/Livecheckables/openssh.rb
+++ b/Livecheckables/openssh.rb
@@ -1,0 +1,6 @@
+class Openssh
+  livecheck do
+    url "https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/"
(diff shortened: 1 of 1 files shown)


Livecheckables/openssh.rb
```diff
@@ -0,0 +1,6 @@
+class Openssh
+  livecheck do
+    url "https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/"
+    regex(/href=.*?openssh-v?(\d+(?:\.\d+)+(?:p\d+))\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?openssh-v?(\d+(?:\.\d+)+(?:p\d+)?)\.t/i)
```

Though the current file names all contain something like `p1` at the end, we may want to make it optional in case we ever encounter a version without it. Looks good outside of this.

nandahkrishna:
Updated 👍🏼

PR #802, opened 2020-06-03
sdl: add livecheckable

nandahkrishna:
They do have a release page at https://www.libsdl.org/release/ but I used http://hg.libsdl.org/SDL/tags instead, to check the tags. An issue (?) I noticed when browsing through the first URL was that their 2+ versions are released as `SDL2-x.x.x` whereas their latest 1.x version, `1.2.5` (homebrew-core has this) is released as just `SDL-1.2.5`. Do let me know if I've got to change something here.

### DIFF
--- a/Livecheckables/sdl.rb
+++ b/Livecheckables/sdl.rb
@@ -1,0 +1,6 @@
+class Sdl
+  livecheck do
+    url "https://www.libsdl.org/release/"
+    regex(/href=.*?SDL-v?(1(?:\.\d+)+)\.t/i)
+  end
+end


Livecheckables/sdl.rb
```diff
@@ -0,0 +1,6 @@
+class Sdl
+  livecheck do
+    url "http://hg.libsdl.org/SDL/tags"
+    regex(/release-v?(\d+(?:\.\d+)+)/i)
```
samford, line 4:
```suggestion
    url "https://www.libsdl.org/release/"
    regex(/href=.*?SDL-v?(1(?:\.\d+)+)\.t/i)
```

I would prefer to stick to the first-party release index, since it's where the stable archive is found. The `sdl` formula is only concerned with `1.x` releases (`2.x` releases are found in the `sdl2` formula), so we just have to make sure to only match appropriate versions.

There are two ways to accomplish this in the context of `sdl`. The file names consistently use `SDL-x.x.x` to refer to `1.x` releases and `SDL2-x.x.x` to refer to `2.x` releases. By only matching `SDL-` in the regex, we should only get `1.x` releases.

However, we can go also further and restrict matching to only versions with a major version of `1`, just to be sure (in case someone messes up and uses `SDL-` when they meant `SDL2-`). This is how we usually restrict matching for versioned formulae when the software name in the file name is the same for all releases and the only distinction is the major version.

nandahkrishna:
Updated 👍🏼

PR #803, opened 2020-06-03
ldns: add livecheckable

nandahkrishna:
Livecheckable for `ldns`.

### DIFF
--- a/Livecheckables/ldns.rb
+++ b/Livecheckables/ldns.rb
@@ -1,0 +1,9 @@
+# We check the GitHub repo tags instead of
+# https://nlnetlabs.nl/downloads/ldns/ since the first-party site has a
+# tendency to lead to an `execution expired` error.
+class Ldns
+  livecheck do
+    url "https://github.com/NLnetLabs/ldns.git"
+    regex(/^(?:release-)?v?(\d+(?:\.\d+)+)$/i)
+  end
+end


Livecheckables/ldns.rb
```diff
@@ -0,0 +1,5 @@
+class Ldns
+  livecheck do
+    url "https://github.com/NLnetLabs/ldns.git"
+  end
```
samford, line 4:
```suggestion
  # We check the GitHub repo tags instead of
  # https://nlnetlabs.nl/downloads/ldns/ since the first-party site has a
  # tendency to lead to an `execution expired` error.
  livecheck do
    url "https://github.com/NLnetLabs/ldns.git"
    regex(/^(?:release-)?v?(\d+(?:\.\d+)+)$/i)
  end
```

It's good to leave a comment here to explain why we use the Git repo for NLnetLabs software (like we do in the livecheckables for `nsd` and `unbound`), otherwise there's a chance that someone could naively modify the livecheckable to use the first-party downloads index instead.

Though this works without a regex at this moment, we need to add one to restrict matching to stable versions. Without a regex, pre-release versions (e.g., `1.7.1-rc1`) show up in the matches and we don't want that.

To be clear, the pre-release version filtering in the heuristic is only used when a livecheckable doesn't exist for the formula, so a regex is required to ensure this works check properly.

Past that, I made the leading `release-` part optional, in case they ever move to tags like `1.2.3` in the future.

nandahkrishna:
Updated 👍🏼

PR #819, opened 2020-06-10
Move/refactor Formula methods

samford:
The `#all_urls` and `#latest` methods in livecheck's Formula extension aren't likely to survive in their current state when livecheck is integrated with Homebrew/brew. That is to say, these methods are only currently used by livecheck and may not be valuable additions to Formula proper.

`#all_urls` is pretty specific to livecheck (it's not actually _all_ URLs in the formula, it's the URLs that are useful for _livecheck_), so it makes sense to move it into `utils.rb`.

`#latest` may or may not end up being useful for other parts of Homebrew/brew in the long run but until it would be useful elsewhere, I've merged it with `version_heuristic` as a `latest_version` function.

This may not be the final shape of these changes, as we'll still have to decide what to do with what's in `heuristic.rb` as part of the migration. This is one step in the path taking us to merging with Homebrew/brew.

---

After this is merged, we'll be in a position where we could get rid of livecheck's `Formula` extension. The only difference at that point will be the conditions used for `return @livecheck if livecheckable? || !block_given?` (compared to the current `return @livecheck unless block_given?`).

This change may need to be incorporated into Homebrew/brew's `Formula.rb`, since it prevents the `livecheck` object from being redefined. Other methods in `Formula` simply use `return something unless block_given?`, so it may be fine to allow for redefinition and we can just remove livecheck's `Formula` extension without integrating the change.


samford:
> Just a question, unless we want to support having both Livecheckables and `livecheck` blocks simultaneously, we wouldn't require the `if livecheckable?` check, right? At any point in time we'll be having only one of the two, so wouldn't it be okay to just remove `livecheck.rb` and not integrate those changes in Homebrew/brew @samford?

Ah, yeah, I had forgotten what led to the addition of the `livecheckable?` condition in the early return. With that in mind, I've updated this to simply remove the `Formula` extension and there's no need to incorporate the `livecheckable?` condition in Homebrew/brew's `Formula#livecheck` method.

PR #822, opened 2020-06-12
fltk: add livecheckable

nandahkrishna:
Adding a Livecheckable for `fltk`.

### DIFF
--- a/Livecheckables/fltk.rb
+++ b/Livecheckables/fltk.rb
@@ -1,0 +1,6 @@
+class Fltk
+  livecheck do
+    url "https://www.fltk.org/software.php"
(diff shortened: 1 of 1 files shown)


Livecheckables/fltk.rb
```diff
@@ -0,0 +1,6 @@
+class Fltk
+  livecheck do
+    url "https://www.fltk.org/software.php"
+    regex(/fltk-v?(\d+(?:\.\d+)+)-source\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?fltk-v?(\d+(?:\.\d+)+(?:-\d+)?)-source\.t/i)
```

We should restrict this to match within `href` attributes, as per usual.

Past that, we need to also match versions like `1.2.3-4`, since this livecheckable is omitting versions `1.3.4-1` and `1.3.4-2 `.

PR #829, opened 2020-06-12
pstree: add livecheckable

nandahkrishna:
The website http://www.thp.uni-duisburg.de/pstree/ is very minimal and the source file linked to doesn't have a version in the name. The alternative (which I've used) is to check the README at http://www.thp.uni-duisburg.de/pstree/README which lists the correct version.

Maybe not the best way to deal with this though, any suggestions?

### DIFF
--- a/Livecheckables/pstree.rb
+++ b/Livecheckables/pstree.rb
@@ -1,0 +1,6 @@
+class Pstree
+  livecheck do
+    url "http://www.thp.uni-duisburg.de/pstree/README"
+    regex(/This is pstree V (\d+(?:\.\d+)+)/i)
+  end
+end


samford:
The text for the README may change in the future, so this doesn't seem like a reliable check to me.

The only alternative when it comes to first-party sources seems to be checking the FTP files. I put together a strategy for checking files in an FTP directory and it works as expected but I'm not yet sure if this would be the way to go. I'll come back to this one.

samford:
Maxim recently brought up the idea of an FTP strategy as well, so I'm considering adding it in the future but I want to discuss how to approach it. He and I have different ideas on how it should be implemented and we have to resolve that before moving forward.

We won't get to this discussion and implementation of the FTP strategy until after the Homebrew/brew migration, so I'm going to close this PR for now. I have a list of formulae that could/would benefit from an FTP strategy (this formula is among them) and I'll be sure to address them if we end up moving forward with an FTP strategy in the future.

PR #831, opened 2020-06-12
git-crypt: add livecheckable

nandahkrishna:
Adding Livecheckable for `git-crypt`. Using their GitHub repository rather than the homepage.

### DIFF
--- a/Livecheckables/git-crypt.rb
+++ b/Livecheckables/git-crypt.rb
@@ -1,0 +1,6 @@
+class GitCrypt
+  livecheck do
+    url "https://github.com/AGWA/git-crypt.git"
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/git-crypt.rb
```diff
@@ -0,0 +1,6 @@
+class GitCrypt
+  livecheck do
+    url "https://github.com/AGWA/git-crypt.git"
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url :homepage
    regex(/href=.*?git-crypt-v?(\d+(?:\.\d+)+)\.t/i)
```

Since the stable archive comes from the first-party website, it makes more sense to me to check the homepage (since there isn't an accessible index page of archive files).

The first-party website references their own Git repo (https://www.agwa.name/git/git-crypt.git) and the GitHub repo is a mirror, so I'm not sure it would be appropriate to use the GitHub repo anyway.

PR #833, opened 2020-06-12
sdl_ttf: add livecheckable

nandahkrishna:
Adding Livecheckable for `sdl_ttf`.

### DIFF
--- a/Livecheckables/sdl_ttf.rb
+++ b/Livecheckables/sdl_ttf.rb
@@ -1,0 +1,6 @@
+class SdlTtf
+  livecheck do
+    url "https://www.libsdl.org/projects/SDL_ttf/release"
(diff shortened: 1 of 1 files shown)


Livecheckables/sdl_ttf.rb
```diff
@@ -0,0 +1,6 @@
+class SdlTtf
+  livecheck do
+    url "https://www.libsdl.org/projects/SDL_ttf/release"
+    regex(/href=.*?SDL_ttf-v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?SDL_ttf-v?(1(?:\.\d+)+)\.t/i)
```

It probably won't hurt to restrict this check to major versions of 1, like we did with the other `sdl*` formulae (in relation to the `sdl2` formulae).

PR #836, opened 2020-06-12
phpmyadmin: add livecheckable

nandahkrishna:
Adding Livecheckable for `phpmyadmin`.

### DIFF
--- a/Livecheckables/phpmyadmin.rb
+++ b/Livecheckables/phpmyadmin.rb
@@ -1,6 +1,6 @@
 class Phpmyadmin
   livecheck do
     url "https://www.phpmyadmin.net/files/"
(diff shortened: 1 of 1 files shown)


Livecheckables/phpmyadmin.rb
```diff
@@ -0,0 +1,6 @@
+class Phpmyadmin
+  livecheck do
+    url "https://www.phpmyadmin.net/files/"
+    regex(/href=.*?phpMyAdmin-v?(\d+(?:\.\d+)+)-all-languages\.zip/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?phpMyAdmin-v?(\d+(?:\.\d+)+)-all-languages\.zip[ '">]/i)
```

If we add a delimiter to the end of the URL (`[ '">]`), we can avoid matching the verification files (e.g., `.zip.asc`, `.zip.sha256`), which keeps the debug output a bit cleaner.

PR #837, opened 2020-06-12
nkf: add livecheckable

nandahkrishna:
Adding Livecheckable for `nkf`. Checking the releases page didn't work so I've check the RSS feed instead.

### DIFF
--- a/Livecheckables/nkf.rb
+++ b/Livecheckables/nkf.rb
@@ -1,6 +1,6 @@
 class Nkf
   livecheck do
-    url "https://osdn.net/projects/nkf/releases/rss"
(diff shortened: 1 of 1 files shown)


Livecheckables/nkf.rb
```diff
@@ -0,0 +1,6 @@
+class Nkf
+  livecheck do
+    url "https://osdn.net/projects/nkf/releases/rss"
+    regex(/<osdn:file url=.*?nkf-v?(\d+(?:\.\d+)+(?:[a-z])?)\.t/i)
```
samford, line 4:
```suggestion
    url "https://osdn.net/projects/nkf/releases/"
    regex(%r{=.*?rel/nkf/v?(\d+(?:\.\d+)+[a-z]?)[ '"]}i)
```

PR #838, opened 2020-06-13
qhull: add livecheckable

nandahkrishna:
Adding Livecheckable for `qhull`. I match only the `year.version` versions like `2020.1` (current latest) or `2009.1.1`. Here's their GitHub repo (https://github.com/qhull/qhull), if there are any changes to make, let me know!

### DIFF
--- a/Livecheckables/qhull.rb
+++ b/Livecheckables/qhull.rb
@@ -1,0 +1,6 @@
+class Qhull
+  livecheck do
+    url "https://github.com/qhull/qhull.git"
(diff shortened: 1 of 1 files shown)


Livecheckables/qhull.rb
```diff
@@ -0,0 +1,6 @@
+class Qhull
+  livecheck do
+    url "https://github.com/qhull/qhull.git"
```
samford, line 3:
```suggestion
    url :head
```

PR #854, opened 2020-06-14
visp: add livecheckable

miccal:
Output before:
```
-bash-5.0.17- /Users/miccal (28) [> brew livecheck visp
Error: visp: Unable to get versions
```
Output after:
```
-bash-5.0.17- /Users/miccal (28) [> brew livecheck visp
visp : 3.3.0 ==> 3.3.0
```
Thank you.

### DIFF
--- a/Livecheckables/visp.rb
+++ b/Livecheckables/visp.rb
@@ -1,0 +1,6 @@
+class Visp
+  livecheck do
+    url :homepage
+    regex(/href=.*?visp-v?(\d+(?:\.\d+)+)/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/visp.rb
```diff
@@ -0,0 +1,6 @@
+class Visp
+  livecheck do
+    url :homepage
+    regex(/href=.*?visp-v?(\d+(?:\.\d+)+)/i)
```
samford, line 4:
```suggestion
    url "https://visp.inria.fr/download/"
    regex(/href=.*?visp-v?(\d+(?:\.\d+)+)\.t/i)
```

miccal: Thanks @samford -- I actually had it this way first, but I was not too sure.

PR #855, opened 2020-06-14
neo4j: add livecheckable

nandahkrishna:
Adding a livecheckable for `neo4j`.

### DIFF
--- a/Livecheckables/neo4j.rb
+++ b/Livecheckables/neo4j.rb
@@ -1,0 +1,7 @@
+class Neo4j
+  livecheck do
+    url "https://neo4j.com/download-center/"
+    regex(/href=.*?edition=community[^ '">]+release=v?(\d+(?:\.\d+)+)[& '">]
+          |href=.*?release=v?(\d+(?:\.\d+)+)[^ '">]+edition=community/ix)
+  end
+end


Livecheckables/neo4j.rb
```diff
@@ -0,0 +1,6 @@
+class Neo4j
+  livecheck do
+    url "https://github.com/neo4j/neo4j.git"
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url "https://neo4j.com/download-center/"
    regex(/href=.*?edition=community&(?:amp;)?release=v?(\d+(?:\.\d+)+)[& '">]/i)
```

nandahkrishna:
It's because of the complications involved with the `regex` and the relative ease of using the GitHub repository that I used it instead 😅. I'll work on the regex and see if I can improve it.

samford:
This could be much simpler if we didn't have to be mindful of the capture group but I think we may be stuck with this mess unless you find something simpler that works properly.

```ruby
    regex(
      /href=.*?edition=community[^ '">]+release=v?(\d+(?:\.\d+)+)[& '">]|
      href=.*?release=v?(\d+(?:\.\d+)+)[^ '">]+edition=community/ix,
    )
```

PR #865, opened 2020-06-16
iftop: add livecheckable

nandahkrishna:
Adding Livecheckable for `iftop`.


Livecheckables/iftop.rb
```diff
@@ -0,0 +1,6 @@
+class Iftop
+  livecheck do
```
samford, line 2:
```suggestion
  # We have to allow the regex to match prerelease versions (e.g., 1.0pre4)
  # until there's a new stable version. The newest version was released on
  # 2014-01-19, so it could be a while.
  livecheck do
```

PR #870, opened 2020-06-16
Add completion for livecheck

nandahkrishna:
Adding `bash` and `zsh` completions for `brew livecheck`. Should add `fish` shortly too. The changes here almost resemble those in Homebrew/brew#7748 entirely but with inputs taken from similar PRs in Homebrew/homebrew-services.

To test the `zsh` completion you'll have to use this symlink:
```zsh
ln -s /usr/local/Homebrew/Library/Taps/homebrew/homebrew-livecheck/completions/zsh/_brew_livecheck /usr/local/share/zsh/site-functions/_brew_functions
```

As for `bash`, I just used `source` for the `completions/bash/brew-livecheck` file and tested it.

If the Tap is (re)installed, the symlinks will be created automatically (this will happen only after the PR is merged, obviously).

As for the Homebrew/brew PR, do I let it be or close it and create a new one when it's time to migrate the command?

CC: @samford

### DIFF
--- a/completions/bash/brew-livecheck
+++ b/completions/bash/brew-livecheck
@@ -1,0 +1,20 @@
+_brew_livecheck() {
+  local cur="${COMP_WORDS[COMP_CWORD]}"
+  case "$cur" in
+    -*)
+      __brewcomp "--verbose --quiet --debug --full-name --tap --installed --json --all --newer-only --help"
+      return
+      ;;
+  esac
+  __brew_complete_formulae
+}
+
+_brew_with_livecheck() {
+  if [ $COMP_CWORD -gt 1 ] && [ "${COMP_WORDS[1]}" = "livecheck" ] ; then
+    _brew_livecheck
+    return
+  fi
+  _brew
+}
+
+complete -o bashdefault -o default -F _brew_with_livecheck brew
--- a/completions/zsh/_brew_livecheck
+++ b/completions/zsh/_brew_livecheck
@@ -1,0 +1,19 @@
+#compdef brew
+#autoload
+
+_brew_livecheck() {
+  _arguments \
+    '(--verbose,-v)'{--verbose,-v}'[Make some output more verbose]' \
+    '(--quiet,-q)'{--quiet,-q}'[Suppress any warnings]' \
+    '(--debug,-d)'{--debug,-d}'[Display any debugging information]' \
+    '--full-name[Print formulae with fully-qualified name]' \
+    '--tap[Check the formulae within the given tap, specified as user/repo]' \
+    '--installed[Check formulae that are currently installed]' \
+    '--json[Output information in JSON format]' \
+    '--all[Check all available formulae]' \
+    '--newer-only[Show the latest version only if it is newer than the formula]' \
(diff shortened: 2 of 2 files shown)


samford:
> As for the Homebrew/brew PR, do I let it be or close it and create a new one when it's time to migrate the command?

It would be good to close the Homebrew/brew PR in favor of this one and we'll integrate this when we do the migration in the future. In situations like this, Mike likes to close PRs in Homebrew/brew (with the idea that a new PR could always be opened in the future) rather than keep them around indefinitely.

I didn't have any luck with this from a quick test, so I'll have to tinker with it more tomorrow. Just so I understand properly, should this also autocomplete the flags (like `--verbose`) and not just `brew livecheck`?

nandahkrishna:
Yeah that's right. It should give you the flags if you type in a `-` or a `--`, else it should give you all possible Formulae (with a prompt asking you if you rally want to, haha).

nandahkrishna:
The completions for `fish` don't seem to be included in the homebrew-services repository – they're actually included at the end of `brew.fish` in Homebrew/brew, so I'm skipping `fish` for now. I think it would be better to take care of it when we're doing the migration.

samford:
Hm, I'm on this PR branch and I used the same commands as in your gif but I haven't had any luck:

```
ln -s /usr/local/Homebrew/Library/Taps/homebrew/homebrew-livecheck/completions/zsh/_brew_livecheck /usr/local/share/zsh/site-functions/_brew_livecheck
autoload -U compinit && compinit
[testing brew livecheck autocompletion]
```

For me, `brew livecheck` will autocomplete but I don't get any autocompletion beyond that (e.g., formulae names, flags, etc.). I've confirmed that the `_brew_livecheck` symlink exists in `/usr/local/share/zsh/site-functions/_brew_livecheck` and it points to the correct file. I also tried this in a new terminal window after creating the symlink as well (running `autoload -U compinit && compinit` before testing `brew livecheck` autocompletion). It's a bit baffling.

However, I do get this autocomplete behavior (demonstrated in your gif) if I use your previous Homebrew/brew modification to `completions/zsh/_brew`. The autocompletion for `brew` and `brew cask` also works fine, for what it's worth.

nandahkrishna:
That's strange.  If other completions are fine, then there shouldn't be any changes required with respect to your `zsh` configuration. I'll try checking what the problem might be.

nandahkrishna:
@samford, you might have to delete _all_ your `~/.zcompdump` files (there may be duplicates such as `.zcompdump-5.7.1`) and then running `compinit`. Weirdly enough, even I wasn't able to use the completions outside the repository until I did this. Could you try this and let me know?

samford:
I deleted the two `.zcompdump`-related files in my home directory and it works as expected now.

Funny enough, I did the same thing earlier and afterward everything was working fine and I didn't put it together that deleting `.zcompdump` was what had fixed it, so I proceeded to break it again by making changes, ahah.

Unless you have any additional changes to make to this PR, I'm fine with merging this since I've confirmed that it works.

nandahkrishna:
No further changes 👍🏼

samford:
Thanks for this and for your help in troubleshooting!

PR #882, opened 2020-06-18
hping: add livecheckable to skip

nandahkrishna:
Adding Livecheckable for `hping` to skip livecheck. Details: http://www.hping.org/download.php

### DIFF
--- a/Livecheckables/hping.rb
+++ b/Livecheckables/hping.rb
@@ -1,4 +1,11 @@
 class Hping
+  # The first-party download page (http://www.hping.org/download.php) has
+  # removed links to any archive files, with a notice that Hping is no longer
+  # actively developed. There won't be any new releases and we can't check for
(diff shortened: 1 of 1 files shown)


Livecheckables/hping.rb
```diff
@@ -0,0 +1,5 @@
+class Hping
+  livecheck do
+    skip "No longer actively developed"
```
samford, line 3:
```suggestion
  # The first-party download page (http://www.hping.org/download.php) has
  # removed links to any archive files, with a notice that Hping is no longer
  # actively developed. There won't be any new releases and we can't check for
  # any in this state, so it's appropriate to skip this. If the GitHub repo
  # (https://github.com/antirez/hping) starts creating releases, then it would
  # be appropriate to update this livecheckable but there are no releases at
  # the time of writing this.
  livecheck do
    skip "No longer actively developed"
```

PR #883, opened 2020-06-18
rpm2cpio: add livecheckable

nandahkrishna:
Adding Livecheckable for `rpm2cpio`. In its current form the Livecheckable is probably less than ideal, however I was unable to find alternatives. Any better options?

### DIFF
--- a/Livecheckables/rpm2cpio.rb
+++ b/Livecheckables/rpm2cpio.rb
@@ -1,0 +1,6 @@
+class Rpm2cpio
+  livecheck do
+    url "https://svnweb.freebsd.org/ports/head/archivers/rpm2cpio/Makefile?view=markup"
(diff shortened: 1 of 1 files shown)


Livecheckables/rpm2cpio.rb
```diff
@@ -0,0 +1,6 @@
+class Rpm2cpio
+  livecheck do
+    url "https://svnweb.freebsd.org/ports/head/archivers/rpm2cpio/Makefile?view=markup"
+    regex(%r{PORTVERSION=\s*v?(\d+(?:\.\d+)+)</td>}i)
```
samford, line 4:
```suggestion
    url "https://svnweb.freebsd.org/ports/head/archivers/rpm2cpio/Makefile?view=co"
    regex(/^PORTVERSION=\s*?v?(\d+(?:\.\d+)+)$/i)
```

PR #885, opened 2020-06-18
mit-scheme: add livecheckable

nandahkrishna:
Adding a livecheckable for `mit-scheme`.

### DIFF
--- a/Livecheckables/mit-scheme.rb
+++ b/Livecheckables/mit-scheme.rb
@@ -1,0 +1,6 @@
+class MitScheme
+  livecheck do
+    url :homepage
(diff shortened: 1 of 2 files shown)


Livecheckables/mit-scheme.rb
```diff
@@ -0,0 +1,6 @@
+class MitScheme
+  livecheck do
+    url :homepage
+    regex(/href=.*?mit-scheme-v?(\d+(?:\.\d+)+)-svm1\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?mit-scheme-v?(\d+(?:\.\d+)+)\.t/i)
```

PR #893, opened 2020-06-21
sentencepiece: add livecheckable

chenrui333:
relates to Homebrew/homebrew-core#56614

Update livecheck to exclude `1.0.0` tag

### before the change

```
$ brew livecheck -v sentencepiece
sentencepiece (guessed) : 0.1.91 ==> 1.0.0
```

### after the change

```
$ brew livecheck -v sentencepiece
sentencepiece : 0.1.91 ==> 0.1.92
```

### DIFF
--- a/Livecheckables/sentencepiece.rb
+++ b/Livecheckables/sentencepiece.rb
@@ -1,6 +1,6 @@
 class Sentencepiece
   livecheck do
-    url "https://github.com/google/sentencepiece.git"
-    regex(/^v(\d+(?:\.\d+)+)$/)
+    url "https://github.com/google/sentencepiece/releases/latest"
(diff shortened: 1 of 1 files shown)


Livecheckables/sentencepiece.rb
```diff
@@ -0,0 +1,6 @@
+class Sentencepiece
+  livecheck do
+    url "https://github.com/google/sentencepiece.git"
+    regex(/^v(\d+(?:\.\d+)+)$/)
```
samford, line 4:
```suggestion
    url "https://github.com/google/sentencepiece/releases/latest"
    regex(%r{href=.*?/tag/v?(\d+(?:\.\d+)+)["']}i)
```

chenrui333: yeah, that also works.

PR #895, opened 2020-06-21
awscli@1: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v awscli@1
awscli@1 : versioned
```

### after the change

```
$ brew livecheck -v awscli@1
awscli@1 : 1.18.70 ==> 1.18.84
```

relates to Homebrew/homebrew-core#56618

### DIFF
--- a/Livecheckables/awscli@1.rb
+++ b/Livecheckables/awscli@1.rb
@@ -1,6 +1,6 @@
 class AwscliAT1
   livecheck do
-    url "https://github.com/aws/aws-cli.git"
-    regex(/v?(1(?:\.\d+)+)$/)
+    url :stable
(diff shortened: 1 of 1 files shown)


Livecheckables/awscli@1.rb
```diff
@@ -0,0 +1,6 @@
+class AwscliAT1
+  livecheck do
+    url "https://github.com/aws/aws-cli.git"
+    regex(/v?(1(?:\.\d+)+)$/)
```
samford, line 4:
```suggestion
    url :stable
    regex(/^v?(1(?:\.\d+)+)$/i)
```

chenrui333: I see, that is pretty neat!

PR #896, opened 2020-06-21
erlang@21: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck erlang@21
erlang@21 : versioned
```

### after the change

```
$ brew livecheck -v erlang@21
erlang@21 : 21.3.8.16 ==> 21.3.8.16
```

### DIFF
--- a/Livecheckables/erlang@21.rb
+++ b/Livecheckables/erlang@21.rb
@@ -1,0 +1,6 @@
+class ErlangAT21
+  livecheck do
+    url :stable
+    regex(/^OTP-(21.[0-9.]+)$/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/erlang@21.rb
```diff
@@ -0,0 +1,6 @@
+class ErlangAT21
+  livecheck do
+    url :stable
+    regex(/^OTP-(21.[0-9.]+)$/i)
```
samford, line 4:
```suggestion
    regex(/^OTP-v?(21(?:\.\d+)+)$/i)
```

chenrui333: my bad. logged into my note for the future ref.

PR #897, opened 2020-06-21
erlang@22: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v erlang@22
erlang@22 : versioned
```

### after the change
```
$ brew livecheck -v erlang@22
erlang@22 : 22.3.4.1 ==> 22.3.4.1
```

### DIFF
--- a/Livecheckables/erlang@22.rb
+++ b/Livecheckables/erlang@22.rb
@@ -1,0 +1,6 @@
+class ErlangAT22
+  livecheck do
+    url :stable
+    regex(/^OTP-(22.[0-9.]+)$/i)
+  end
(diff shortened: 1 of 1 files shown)


Livecheckables/erlang@22.rb
```diff
@@ -0,0 +1,6 @@
+class ErlangAT22
+  livecheck do
+    url :stable
+    regex(/^OTP-(22.[0-9.]+)$/i)
```
samford, line 4:
```suggestion
    regex(/^OTP-v?(22(?:\.\d+)+)$/i)
```

PR #898, opened 2020-06-21
workflows: add PR template

chenrui333:
Add a simple PR template to pre-populate the structure.


chenrui333:
That is fine, I can take it.

PR #899, opened 2020-06-21
grails: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v grails
grails (guessed) : 4.0.3 ==> 4.1.0.M1
```

### after the change

```
$ brew livecheck -v grails
grails : 4.0.3 ==> 4.0.3
```

### DIFF
--- a/Livecheckables/grails.rb
+++ b/Livecheckables/grails.rb
@@ -1,0 +1,6 @@
+class Grails
+  livecheck do
+    url :stable
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
+  end
+end


chenrui333:
But that is a milestone release, and I assume we don't ship with milestone release or beta release.

chenrui333:
And to the release notes, it is going to break all the applications with Java 8 (while Java 8 is still in TLS support), we can certainly raise this in the upstream repo though.

samford:
> that is a milestone release

You're right. I've spent too much time looking at different version schemes and I interpreted the "M" as "mark", like one or two formulae have released a `1.2.3` version followed by `1.2.3_mark2` to fix some issue (instead of something reasonable like `1.2.3.1`).

The latest version on the [Grails download page](https://grails.org/download.html) is still 4.0.3, which supports your statement.

PR #900, opened 2020-06-21
epic5: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v epic5
epic5 (guessed) : 2.1.2 ==> 5-2.1.2
```

### after the change

```
$ brew livecheck -v epic5
epic5 : 2.1.2 ==> 2.1.2
```

### DIFF
--- a/Livecheckables/epic5.rb
+++ b/Livecheckables/epic5.rb
@@ -1,6 +1,6 @@
 class Epic5
   livecheck do
-    url :head
-    regex(/^epic5-(\d+(?:\.\d+)+)$/i)
+    url "http://ftp.epicsol.org/pub/epic/EPIC5-PRODUCTION/"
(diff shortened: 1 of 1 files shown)


Livecheckables/epic5.rb
```diff
@@ -0,0 +1,6 @@
+class Epic5
+  livecheck do
+    url :head
+    regex(/^epic5-(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url "http://ftp.epicsol.org/pub/epic/EPIC5-PRODUCTION/"
    regex(/href=.*?epic5-v?(\d+(?:\.\d+)+)\.t/i)
```

chenrui333: Good call!

PR #901, opened 2020-06-21
genders: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v genders
genders (guessed) : 1.27.3 ==> 2-0-1pre
```

### after the change

```
$ brew livecheck -v genders
genders : 1.27.3 ==> 1-27-3
```

### DIFF
--- a/Livecheckables/genders.rb
+++ b/Livecheckables/genders.rb
@@ -1,6 +1,6 @@
 class Genders
   livecheck do
     url "https://github.com/chaos/genders/releases/latest"
-    regex(%r{href=.*?/tag/genders-(\d+(?:-\d+)+)["']}i)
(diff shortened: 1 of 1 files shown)


Livecheckables/genders.rb
```diff
@@ -0,0 +1,6 @@
+class Genders
+  livecheck do
+    url "https://github.com/chaos/genders/releases/latest"
+    regex(%r{href=.*?/tag/genders-(\d+(?:-\d+)+)["']}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?/tag/genders-v?(\d+(?:-\d+)+)["']}i)
```

I still like to have the optional `v?`, just in case.

PR #904, opened 2020-06-21
mesa: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v mesa
mesa (guessed) : 20.1.1 ==> 20090313
```

### after the change

```
$ brew livecheck -v mesa
mesa : 20.1.1 ==> 20.1.1
```

### DIFF
--- a/Livecheckables/mesa.rb
+++ b/Livecheckables/mesa.rb
@@ -1,6 +1,6 @@
 class Mesa
   livecheck do
-    url "https://mesa.freedesktop.org/archive/"
+    url "https://archive.mesa3d.org"
     regex(/href="mesa-(\d+(?:\.\d+)+)\.t/i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/mesa.rb
```diff
@@ -0,0 +1,6 @@
+class Mesa
+  livecheck do
+    url "https://mesa.freedesktop.org/archive/"
```
samford, line 3:
```suggestion
    url "https://archive.mesa3d.org"
```

The above URL redirects to https://archive.mesa3d.org/, so we should use that to avoid the redirection.

Livecheckables/mesa.rb
```diff
@@ -0,0 +1,6 @@
+class Mesa
+  livecheck do
+    url "https://archive.mesa3d.org"
+    regex(/href="mesa-(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?mesa-(\d+(?:\.\d+)+)\.t/i)
```

I missed this on the first pass. We use `href=.*?` these days, as a given HTML page may use `"`, `'`, or even nothing at all (much to my horror).

PR #905, opened 2020-06-21
mavsdk: add livecheckable

chenrui333:
### before

```
$ brew livecheck -v mavsdk
mavsdk (guessed) : 0.27.0 ==> 9
```

### after

```
$ brew livecheck -v mavsdk
mavsdk : 0.27.0 ==> 0.28.0
```

relates to Homebrew/homebrew-core#56620

### DIFF
--- a/Livecheckables/mavsdk.rb
+++ b/Livecheckables/mavsdk.rb
@@ -1,6 +1,6 @@
 class Mavsdk
   livecheck do
-    url :satble
+    url :stable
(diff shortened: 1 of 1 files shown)


Livecheckables/mavsdk.rb
```diff
@@ -0,0 +1,6 @@
+class Mavsdk
+  livecheck do
+    url :satble
```
samford, line 3:
```suggestion
    url :stable
```

PR #906, opened 2020-06-21
ledger: add livecheckable

chenrui333:
### before

```
$ brew livecheck -v ledger
ledger (guessed) : 3.2.1 ==> 2006
```

### after the change

```
$ brew livecheck -v ledger
ledger : 3.2.1 ==> 3.2.1
```

### DIFF
--- a/Livecheckables/ledger.rb
+++ b/Livecheckables/ledger.rb
@@ -1,6 +1,6 @@
 class Ledger
   livecheck do
-    url :stable
+    url :head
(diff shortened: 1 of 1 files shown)


Livecheckables/ledger.rb
```diff
@@ -0,0 +1,6 @@
+class Ledger
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

The `ledger` formula has the GitHub repo as `head`, so we can just use that directly.

PR #908, opened 2020-06-21
sslscan: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v sslscan
sslscan (guessed) : 1.11.13 ==> 1.11.13-rbsec
```

### after the change

```
$ brew livecheck -v sslscan
sslscan : 1.11.13 ==> 1.11.13
```

### DIFF
--- a/Livecheckables/sslscan.rb
+++ b/Livecheckables/sslscan.rb
@@ -1,6 +1,6 @@
 class Sslscan
   livecheck do
     url :stable
-    regex(/^v?(\d+(?:\.\d+)+)-rbsec$/i)
+    regex(/^v?(\d+(?:\.\d+)+)(?:-rbsec)?$/i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/sslscan.rb
```diff
@@ -0,0 +1,6 @@
+class Sslscan
+  livecheck do
+    url :stable
+    regex(/^v?(\d+(?:\.\d+)+)-rbsec$/i)
```
samford, line 4:
```suggestion
    regex(/^v?(\d+(?:\.\d+)+)(?:-rbsec)?$/i)
```

It would probably be good to make the `-rbsec` optional, in case they drop it from tags anytime in the future.

samford, line 4:
```suggestion
    regex(/^v?(\d+(?:\.\d+)+)(?:-rbsec)?$/i)
```

It would probably be good to make the `-rbsec` optional, in case they drop it from tags anytime in the future.

PR #909, opened 2020-06-21
clib: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v clib
clib (guessed) : 1.11.3 ==> 2.0.0-next.12
```

### after the change

```
$ brew livecheck -v clib
clib : 1.11.3 ==> 1.11.3
```

### DIFF
--- a/Livecheckables/clib.rb
+++ b/Livecheckables/clib.rb
@@ -1,0 +1,6 @@
+class Clib
+  livecheck do
+    url :stable
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/clib.rb
```diff
@@ -0,0 +1,6 @@
+class Clib
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

The formula provides the GitHub repo as the `head` URL, so we can simply use that instead of `:stable` + URL preprocessing.

PR #911, opened 2020-06-21
bison@2.7: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v bison@2.7
bison@2.7 : versioned
```

### after the change

```
$ brew livecheck -v bison@2.7
bison@2.7 : 2.7.1 ==> 2.7.1
```

### DIFF
--- a/Livecheckables/bison@2.7.rb
+++ b/Livecheckables/bison@2.7.rb
@@ -1,0 +1,6 @@
+class BisonAT27
+  livecheck do
+    url "https://ftp.gnu.org/gnu/bison/?C=M&O=D"
+    regex(%r{href="bison-(2(?:\.\d+)+)\.t}i)
+  end
+end


nandahkrishna:
The style check fails here because we need the regex to be enclosed within `//` and not `%r{}` (as there aren't any `/` within the regex itself).

Livecheckables/bison@2.7.rb
```diff
@@ -0,0 +1,6 @@
+class BisonAT27
+  livecheck do
+    url "https://ftp.gnu.org/gnu/bison/?C=M&O=D"
+    regex(%r{href="bison-(2(?:\.\d+)+)\.t}i)
```
nandahkrishna, line 4:
```suggestion
    regex(/href=.*?bison-(2(?:\.\d+)+)\.t/i)
```
Just a doubt as well, the formula name seems to be `bison@2.7` so would we want to match only `2.7.x` or all `2.x.x`? The current regex matches all `2.x.x`.

PR #913, opened 2020-06-21
cgns: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v cgns
cgns (guessed) : 4.1.1 ==> 40
```

### after the change

```
$ brew livecheck -v cgns
cgns : 4.1.1 ==> 4.1.1
```

### DIFF
--- a/Livecheckables/cgns.rb
+++ b/Livecheckables/cgns.rb
@@ -1,6 +1,6 @@
 class Cgns
   livecheck do
-    url :stable
+    url :head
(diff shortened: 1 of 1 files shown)


Livecheckables/cgns.rb
```diff
@@ -0,0 +1,6 @@
+class Cgns
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

PR #914, opened 2020-06-21
chocolate-doom: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v chocolate-doom
chocolate-doom (guessed) : 3.0.0 ==> 5
```

### after the change

```
$ brew livecheck -v chocolate-doom
chocolate-doom : 3.0.0 ==> 3.0.0
```

### DIFF
--- a/Livecheckables/chocolate-doom.rb
+++ b/Livecheckables/chocolate-doom.rb
@@ -1,6 +1,6 @@
 class ChocolateDoom
   livecheck do
-    url :head
-    regex(/^chocolate-doom-(\d+(?:\.\d+)+)$/i)
+    url "https://www.chocolate-doom.org/downloads/"
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?[ '">]}i)
   end
 end
--- a/Livecheckables/chocolate-doom.rb
+++ b/Livecheckables/chocolate-doom.rb
@@ -1,0 +1,6 @@
+class ChocolateDoom
+  livecheck do
+    url :head
+    regex(/^chocolate-doom-(\d+(?:\.\d+)+)$/i)
+  end
+end


Livecheckables/chocolate-doom.rb
```diff
@@ -0,0 +1,6 @@
+class ChocolateDoom
+  livecheck do
+    url :head
+    regex(/^chocolate-doom-(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url "https://www.chocolate-doom.org/downloads/"
    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?[ '">]}i)
```

chenrui333: can you educate me the difference with or without `/?[ '">]`?

samford: Sure, it requires a bit of explaining. For context, the `href` attribute on the page is like `href="1.2.3/"`.

As we've discussed before, we almost always need some sort of delimiter after the numeric part (e.g., `1.2.3`) to avoid matching unstable/undesirable versions (e.g., `1.2.3-rc1`).

With Git tags, the `$` at the end of the regex targets the end of the string and ensures nothing comes after the numeric part. With archive file URLs, we usually use the file extension (like `\.t`, `\.z`, etc.) for this purpose.

When we don't have a file extension, we have to use something else as the ending delimiter. If the trailing forward slash was guaranteed always exist until the end of time, we could use that. However, I don't like to rely on the trailing forward slash being present (it could easily change), so I make it optional.

With an optional forward slash, I use the end of the `href` attribute or `a` tag as the ending delimiter and that's what the `[ '">]` part is all about. The `href` attribute may end in a `"`, a `'`, or nothing at all. If there's another attribute after it, there would be a space in between, at the very least. If there's no other attribute, then we can rely on there being a right angle bracket (`>`) to close the `a` tag.

To summarize, `/?` is an optional forward slash after the numeric part and `[ '">]` matches the end of the `href` attribute or `a` tag, acting as a trailing delimiter to avoid unstable versions. This means we match URLs like `1.2.3/` and `1.2.3` but not `1.2.3-rc1/` or `1.2.3-rc1`.

Does this make sense?

chenrui333: yeah, make sense. Thanks for the detailed explanations!

PR #915, opened 2020-06-21
tcping: add livecheckable

nandahkrishna:
Adding Livecheckable for `tcping`.

### DIFF
--- a/Livecheckables/tcping.rb
+++ b/Livecheckables/tcping.rb
@@ -1,0 +1,6 @@
+class Tcping
+  livecheck do
+    url "https://github.com/mkirchner/tcping.git"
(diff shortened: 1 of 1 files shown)


nandahkrishna:
We use a Gentoo mirror URL to get the Formula's source, and I found this GitHub repository from Gentoo's listing of this package. Is it appropriate for us to check this: https://github.com/mkirchner/tcping @samford?

Livecheckables/tcping.rb
```diff
@@ -0,0 +1,6 @@
+class Tcping
+  livecheck do
+    url "https://github.com/mkirchner/tcping.git"
```
samford, line 3:
```suggestion
    url :head
```

PR #920, opened 2020-06-21
markdown: add livecheckable

nandahkrishna:
Adding Livecheckable for `markdown`.

### DIFF
--- a/Livecheckables/markdown.rb
+++ b/Livecheckables/markdown.rb
@@ -1,0 +1,6 @@
+class Markdown
+  livecheck do
+    url :homepage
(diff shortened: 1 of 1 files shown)


Livecheckables/markdown.rb
```diff
@@ -0,0 +1,6 @@
+class Markdown
+  livecheck do
+    url :homepage
+    regex(/href=.*?Markdown_v?(\d+(?:\.\d+)+)\.zip/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?Markdown.v?(\d+(?:\.\d+)+)\.zip/i)
```

It may be better to match anything here instead of `_`, so this would still match the archive if it was named `Markdown-1.2.3.zip`, for instance.

PR #922, opened 2020-06-21
pidof: add livecheckable

nandahkrishna:
Adding Livecheckable for `pidof`.

### DIFF
--- a/Livecheckables/pidof.rb
+++ b/Livecheckables/pidof.rb
@@ -1,0 +1,6 @@
+class Pidof
+  livecheck do
+    url :homepage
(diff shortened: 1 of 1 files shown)


Livecheckables/pidof.rb
```diff
@@ -0,0 +1,6 @@
+class Pidof
+  livecheck do
+    url :homepage
+    regex(/href=.*?pidof.*Download \(v?(\d+(?:\.\d+)+)\)</i)
```
samford, line 4:
```suggestion
    regex(/href=.*?pidof[^>]+>\s*Download \(v?(\d+(?:\.\d+)+)\)</i)
```

PR #928, opened 2020-06-21
jhead: add livecheckable

nandahkrishna:
Adding Livecheckable for `jhead`.


Livecheckables/jhead.rb
```diff
@@ -0,0 +1,6 @@
+class Jhead
+  livecheck do
+    url "https://www.sentex.net/~mwandel/jhead/"
```
samford, line 3:
```suggestion
    url :homepage
```

nandahkrishna: Oops, didn't realise I just copy-pasted the URL.

PR #929, opened 2020-06-21
spidermonkey: add livecheckable

nandahkrishna:
Adding Livecheckable for `spidermonkey`.

### DIFF
--- a/Livecheckables/spidermonkey.rb
+++ b/Livecheckables/spidermonkey.rb
@@ -1,6 +1,6 @@
 class Spidermonkey
   livecheck do
     url "https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Releases"
-    regex(%r{href=.*?Releases/v?(\d+(?:\.\d+)*)}i)
+    regex(%r{href=.*?Releases/v?(\d+(?:\.\d+)*)[ '">]}i)
   end
 end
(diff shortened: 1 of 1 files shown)


nandahkrishna:
Note from website:
```
Note: Standalone SpiderMonkey is not an official product. Our continuous integration system does produce a tarball that is built into a binary that runs our smoke tests, but we do not maintain it nor actively test its suitability for general embedding. We have periodically created "releases", but they are best-effort and incomplete. We do happily accept patches, and make some effort to keep the tip of the Gecko tree minimally working as an embeddable source package. We are very limited in our ability to support older versions, including those labeled as "releases" on this page.
```
Not sure if we should be skipping this. Also, their release system is confusing – apparently `1.8.8` was not actually "released", and after that version they started using version numbers like `17`, `24` etc.

Livecheckables/spidermonkey.rb
```diff
@@ -0,0 +1,6 @@
+class Spidermonkey
+  livecheck do
+    url "https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Releases"
+    regex(%r{href=.*?Releases/v?(\d+(?:\.\d+)*)}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?Releases/v?(\d+(?:\.\d+)*)[ '">]}i)
```

This could use a delimiter, to deal with the very unlikely case that they post an unstable release at some point.

samford:
I was researching this before looking at your livecheckable and I ended up settling on the same page. This is a reasonable approach to me, considering the unusual upstream situation (with regard to being able to identify new versions).

PR #930, opened 2020-06-21
sleepwatcher: add livecheckable

nandahkrishna:
Adding Livecheckable for `sleepwatcher`.

### DIFF
--- a/Livecheckables/sleepwatcher.rb
+++ b/Livecheckables/sleepwatcher.rb
@@ -1,0 +1,6 @@
+class Sleepwatcher
+  livecheck do
+    url :homepage
+    regex(/href=.*?sleepwatcher_v?(\d+(?:\.\d+)+)\.t/i)
+  end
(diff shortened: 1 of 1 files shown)


nandahkrishna:
We're now into the sub-5k set of Formulae. Of course, this is excluding versioned Formulae and some other cases like those Formula for which livecheck guesses incorrect versions.

Livecheckables/sleepwatcher.rb
```diff
@@ -0,0 +1,6 @@
+class Sleepwatcher
+  livecheck do
+    url :homepage
+    regex(/href=.*?sleepwatcher_v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?sleepwatcher.v?(\d+(?:\.\d+)+)\.t/i)
```

PR #931, opened 2020-06-21
pdftk-java: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v pdftk-java
pdftk-java (guessed) : 3.1.2 ==> 52
```

### after the change

```
$ brew livecheck -v pdftk-java
pdftk-java : 3.1.2 ==> 3.1.2
```

### DIFF
--- a/Livecheckables/pdftk-java.rb
+++ b/Livecheckables/pdftk-java.rb
@@ -1,0 +1,6 @@
+class PdftkJava
+  livecheck do
+    url :stable
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/pdftk-java.rb
```diff
@@ -0,0 +1,6 @@
+class PdftkJava
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

The stable URL is a GitLab archive, which gets converted into the Git repo URL, so it's the same as `:head`. As we discussed, it's better to be explicit.

PR #932, opened 2020-06-21
pipes-sh: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v pipes-sh
pipes-sh (guessed) : 1.3.0 ==> 2014-02-21
```

### after the change

```
$ brew livecheck -v pipes-sh
pipes-sh : 1.3.0 ==> 1.3.0
```

### DIFF
--- a/Livecheckables/pipes-sh.rb
+++ b/Livecheckables/pipes-sh.rb
@@ -1,6 +1,6 @@
 class PipesSh
   livecheck do
-    url :stable
+    url :head
     regex(/^v?(\d+(?:\.\d+)+)$/i)
   end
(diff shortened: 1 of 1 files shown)


Livecheckables/pipes-sh.rb
```diff
@@ -0,0 +1,6 @@
+class PipesSh
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

The stable URL is a GitHub archive, which gets converted into the Git repo URL and is identical to the formula's `head` URL.

PR #933, opened 2020-06-21
plank: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v plank
plank (guessed) : 1.6 ==> 100
```

### after the change

```
$ brew livecheck -v plank
plank : 1.6 ==> 1.6
```

### DIFF
--- a/Livecheckables/plank.rb
+++ b/Livecheckables/plank.rb
@@ -1,0 +1,6 @@
+class Plank
+  livecheck do
+    url :stable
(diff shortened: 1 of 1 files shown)


Livecheckables/plank.rb
```diff
@@ -0,0 +1,6 @@
+class Plank
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :head
```

PR #935, opened 2020-06-21
jdnssec-tools: update url and regex

chenrui333:
### before the change

```
$ brew livecheck -v jdnssec-tools
jdnssec-tools : 0.15 ==> 0.16
```

### after the change

```
$ brew livecheck -v jdnssec-tools
jdnssec-tools : 0.15 ==> 0.15
```

relates to Homebrew/homebrew-core#56153

### DIFF
--- a/Livecheckables/jdnssec-tools.rb
+++ b/Livecheckables/jdnssec-tools.rb
@@ -1,6 +1,6 @@
 class JdnssecTools
   livecheck do
     url "https://github.com/dblacka/jdnssec-tools/releases/latest"
-    regex(%r{href=.+?/jdnssec-tools-v?(\d+(?:\.\d+)+)\.t}i)
+    regex(%r{href=.*?/tag/v?(\d+(?:\.\d+)+)["']}i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/jdnssec-tools.rb
```diff
@@ -1,6 +1,6 @@
 class JdnssecTools
   livecheck do
-    url :head
-    regex(/^v?(\d+(?:\.\d+)+)$/)
+    url "https://github.com/dblacka/jdnssec-tools/releases/latest"
+    regex(%r{href=.+?/jdnssec-tools-v?(\d+(?:\.\d+)+)\.t}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?/tag/v?(\d+(?:\.\d+)+)["']}i)
```

An archive with this type of name may or may not be present in a given release, so it's more reliable to match the tag when checking the "latest" release page.

chenrui333: I am actually debating on relying on the latest release (in the past, we did see the latest does not include the needed artifact either).

But I guess we can turn it into a upstream repo issue instead.

PR #937, opened 2020-06-21
kamel: add livecheckable

chenrui333:
### before the change

```
$ brew livecheck -v kamel
kamel (guessed) : 1.0.0 ==> 1.0.0-nightly.202005130003
```

### after the change

```
$ brew livecheck -v kamel
kamel : 1.0.0 ==> 1.0.0
```

### DIFF
--- a/Livecheckables/kamel.rb
+++ b/Livecheckables/kamel.rb
@@ -1,0 +1,6 @@
+class Kamel
+  livecheck do
+    url :stable
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/kamel.rb
```diff
@@ -0,0 +1,6 @@
+class Kamel
+  livecheck do
+    url :stable
```
samford, line 3:
```suggestion
    url :stable
```

The stable and `head` URLs are both the Git repo in this context but it's probably a _little_ more appropriate to use `:head` here, as the stable URL could change to something different.

PR #945, opened 2020-06-22
bsdiff: add livecheckable

nandahkrishna:
Adding Livecheckable for `bsdiff`.


nandahkrishna:
The homepage uses `https` but that redirects to `http`.

Livecheckables/bsdiff.rb
```diff
@@ -0,0 +1,6 @@
+class Bsdiff
+  livecheck do
```
samford, line 2:
```suggestion
  # We use the HTTP URL for the homepage here because the site redirects this
  # from HTTPS to HTTP, which makes open-uri give a "redirection forbidden"
  # error. The site supports HTTPS, it just doesn't want to use it for the
  # bsdiff page, for whatever reason.
  livecheck do
```

PR #946, opened 2020-06-22
pip-completion: add livecheckable

nandahkrishna:
Adding Livecheckable for `pip-completion`.

### DIFF
--- a/Livecheckables/pip-completion.rb
+++ b/Livecheckables/pip-completion.rb
@@ -1,0 +1,9 @@
+class PipCompletion
+  # The Formula's GitHub repo at https://github.com/ekalinin/pip-bash-completion
+  # does not have any releases or tags, so it isn't possible to check for newer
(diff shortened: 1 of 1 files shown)


Livecheckables/pip-completion.rb
```diff
@@ -0,0 +1,9 @@
+class PipCompletion
+  # The Formula's GitHub repo at https://github.com/ekalinin/pip-bash-completion
+  # does not have any releases or tags, so it isn't possible to check for newer
+  # versions. The Formula definition uses a 'YYYYMMDD' timestamp to indicate the
+  # version.
```
samford, line 5:
```suggestion
  # There currently aren't any versions of pip-completion and the formula
  # simply uses a revision from the upstream GitHub repo. The YYYYMMDD version
  # in the formula isn't from upstream and was created on our end to indicate
  # the date of the revision that's being used.
```

Your text was good but I reworked this to explain things a little differently.

PR #947, opened 2020-06-22
nexus: add livecheckable

nandahkrishna:
Adding Livecheckable for `nexus`.


nandahkrishna:
Oh I see, didn't realise it 😅

PR #949, opened 2020-06-23
perl: update livecheckable

miccal:
Output before:
```
-bash-5.0.17- /Users/miccal (30) [> brew livecheck perl
perl : 5.30.3 ==> 5.30.3
```
Output after:
```
-bash-5.0.17- /Users/miccal (30) [> brew livecheck perl
perl : 5.30.3 ==> 5.32.0
```
I noticed that `livecheck` was not reporting the [latest version](https://github.com/Homebrew/homebrew-core/pull/56701), and I have tried to follow the strategy used in recent PR's.

Thank you (as always).

### DIFF
--- a/Livecheckables/perl.rb
+++ b/Livecheckables/perl.rb
@@ -1,6 +1,6 @@
 class Perl
   livecheck do
-    url "https://www.perl.org/get.html"
-    regex(%r{href=".*?/perl-([0-9.]+)\.t})
+    url :head
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
   end
 end
--- a/Livecheckables/perl.rb
+++ b/Livecheckables/perl.rb
@@ -1,6 +1,6 @@
 class Perl
   livecheck do
-    url :head
-    regex(/^v?(\d+(?:\.\d+)+)$/i)
+    url "https://www.cpan.org/src/"
+    regex(/href=.*?perl-v?(\d+(?:\.\d+)+)\.t/)
   end
 end


Livecheckables/perl.rb
```diff
@@ -1,6 +1,6 @@
 class Perl
   livecheck do
-    url "https://www.perl.org/get.html"
-    regex(%r{href=".*?/perl-([0-9.]+)\.t})
+    url :head
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    url "https://www.cpan.org/src/"
    regex(/href=.*?perl-v?(\d+(?:\.\d+)+)\.t/)
```

miccal:
Thanks @samoford -- I was debating wether to use `:head` or the relevant part of `:stable` for this one.

samford:
> I was debating wether to use `:head` or the relevant part of `:stable` for this one

I generally try to create a check using the stable source first. The first thing I usually do is copy the URL leading up to the file name and see if it's an index page that we can check. If there's some issue with this approach (e.g., it's not an index page, file listings are forbidden, etc.), we sometimes check the first-party site, if that's a good alternative.

We primarily check `:head` when the stable URL comes from the same source as the repo (e.g., a GitHub or GitLab archive) or as an alternative when we can't create a good or reliable check using something close to the stable archive. As an aside, sometimes we check the "latest" release on GitHub instead because some upstream repos will create a tag for a version before it's a real release (so checking the "latest" release avoids the issue).

I'm hoping to write this kind of information up in documentation, however rough, in the near future. Some other maintainers reminded me that it would be difficult for anyone to take over if I wasn't able to fulfill my duties (primarily related to our ongoing Google Summer of Code project), so I'm hoping to get the information out of my brain and in some readable form soon. It should be beneficial to anyone who writes livecheckables as well. The first draft may be a bit rough, due to time constraints, but I'm planning to create comprehensive documentation for livecheck eventually.

Anyway, thanks @miccal!

miccal:
You're welcome @samford.

PR #963, opened 2020-06-23
grace: add livecheckable

nandahkrishna:
Adding Livecheckable for `grace`. The `Download` page inked to in the `homepage` is actually an FTP directory, so unless we have the strategy for that, we should check the Debian archives (which we use).

### DIFF
--- a/Livecheckables/grace.rb
+++ b/Livecheckables/grace.rb
@@ -1,6 +1,6 @@
 class Grace
   livecheck do
     url "https://deb.debian.org/debian/pool/main/g/grace/"
-    regex(/href=.*?grace_v?(\d+(?:\.\d+)+)\.orig\.t/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/grace.rb
```diff
@@ -0,0 +1,6 @@
+class Grace
+  livecheck do
+    url "https://deb.debian.org/debian/pool/main/g/grace/"
+    regex(/href=.*?grace_v?(\d+(?:\.\d+)+)\.orig\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?grace.v?(\d+(?:\.\d+)+)\.orig\.t/i)
```

samford:
I'm still debating whether an FTP strategy is a good idea (re: `pstree`). I've written the code and it works but I'm still not totally told (since it's more intensive than fetching a web page, Git tags, etc.).

I'm fine with checking Debian when the stable archive comes from there and we don't have a good first-party source to check (as here).

PR #965, opened 2020-06-23
ncview: add livecheckable

nandahkrishna:
Adding Livecheckable for `ncview`.

### DIFF
--- a/Livecheckables/ncview.rb
+++ b/Livecheckables/ncview.rb
@@ -1,6 +1,10 @@
 class Ncview
+  # The stable archive in the formula is fetched over FTP and the website for
+  # the software hasn't been updated to list the latest release (it has been
+  # years now). We're checking Debian for now because it's potentially better
+  # than having no check at all.
   livecheck do
     url "http://ftp.debian.org/debian/pool/main/n/ncview/"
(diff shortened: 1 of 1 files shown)


nandahkrishna:
Sadly, the homepage(s) for this project either list an older version (`2.1.7`) to be the latest, or don't list it at all. I've used the Debian listing for this Livecheckable, but we could find alternatives.

Livecheckables/ncview.rb
```diff
@@ -0,0 +1,6 @@
+class Ncview
+  livecheck do
+    url "http://ftp.debian.org/debian/pool/main/n/ncview/"
+    regex(/href=.*?ncview_v?(\d+(?:\.\d+)+)\+ds\.orig\.t/i)
```
samford, line 4:
```suggestion
  # The stable archive in the formula is fetched over FTP and the website for
  # the software hasn't been updated to list the latest release (it has been
  # years now). We're checking Debian for now because it's potentially better
  # than having no check at all.
  livecheck do
    url "http://ftp.debian.org/debian/pool/main/n/ncview/"
    regex(/href=.*?ncview_v?(\d+(?:\.\d+)+)(?:\+ds)?\.orig\.t/i)
```

PR #966, opened 2020-06-23
libsvm: add livecheckable

nandahkrishna:
Adding Livecheckable for `libsvm`. Their website has a download link but that does not have the version number in it, and I could only create a Livecheckable by matching text. They have a GitHub repo but the versioning there is done without the `.`s, i.e., `v324` instead of `v3.2.4`.

### DIFF
--- a/Livecheckables/libsvm.rb
+++ b/Livecheckables/libsvm.rb
@@ -1,6 +1,6 @@
 class Libsvm
   livecheck do
     url :homepage
-    regex(/The current release \(Version v?(\d+(?:\.\d+)+),/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/libsvm.rb
```diff
@@ -0,0 +1,6 @@
+class Libsvm
+  livecheck do
+    url :homepage
+    regex(/The current release \(Version v?(\d+(?:\.\d+)+),/i)
```
samford, line 4:
```suggestion
    regex(/The current release \(Version v?(\d+(?:\.\d+)+)[, )]/i)
```

This is just a small tweak to allow either a comma, space, or closing parenthesis as the trailing delimiter, in case the format changes slightly.

PR #968, opened 2020-06-23
ntp: add livecheckable

nandahkrishna:
Adding Livecheckable for `ntp`.


nandahkrishna:
The project plans to change its versioning scheme soon and this Livecheckable should be updated when that happens. Link: http://support.ntp.org/bin/view/Main/ReleaseNumberingScheme

PR #973, opened 2020-06-23
tcptrace: add livecheckable

nandahkrishna:
Adding Livecheckable for `tcptrace`. The homebrew-core Formula definition for `tcptrace` has a comment stating that the website "is currently offline", but that doesn't seem to be the case.

### DIFF
--- a/Livecheckables/tcptrace.rb
+++ b/Livecheckables/tcptrace.rb
@@ -1,0 +1,6 @@
+class Tcptrace
+  livecheck do
+    url "http://www.tcptrace.org/download.shtml"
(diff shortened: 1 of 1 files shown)


Livecheckables/tcptrace.rb
```diff
@@ -0,0 +1,6 @@
+class Tcptrace
+  livecheck do
```
samford, line 2:
```suggestion
  # This site has a history of going down for periods of time, which is why the
  # formula uses mirrors. This is something to be aware of if livecheck is
  # unable to find versions.
  livecheck do
```

PR #983, opened 2020-06-23
libcuefile: add livecheckable

nandahkrishna:
Adding Livecheckable for `libcuefile`.

### DIFF
--- a/Livecheckables/libcuefile.rb
+++ b/Livecheckables/libcuefile.rb
@@ -1,0 +1,6 @@
+class Libcuefile
+  livecheck do
+    url "https://www.musepack.net/index.php?pg=src"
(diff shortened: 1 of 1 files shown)


Livecheckables/libcuefile.rb
```diff
@@ -0,0 +1,6 @@
+class Libcuefile
+  livecheck do
+    url "https://www.musepack.net/index.php?pg=src"
+    regex(/href=.*?libcuefile[._-]v?(r\d+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?libcuefile[._-](r\d+)\.t/i)
```

Based on the file name format, I don't imagine we'll encounter a file name like `libcuefile_vr123.tar.gz`. I'm fine with omitting the`v?` here, as it doesn't use a common version scheme.

PR #984, opened 2020-06-23
musepack: add livecheckable

nandahkrishna:
Adding Livecheckable for `musepack`.

### DIFF
--- a/Livecheckables/musepack.rb
+++ b/Livecheckables/musepack.rb
@@ -1,6 +1,6 @@
 class Musepack
   livecheck do
     url "https://www.musepack.net/index.php?pg=src"
-    regex(/href=.*?musepack[._-]src[._-]v?(r\d+)\.t/i)
+    regex(/href=.*?musepack[._-]src[._-](r\d+)\.t/i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/musepack.rb
```diff
@@ -0,0 +1,6 @@
+class Musepack
+  livecheck do
+    url "https://www.musepack.net/index.php?pg=src"
+    regex(/href=.*?musepack[._-]src[._-]v?(r\d+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?musepack[._-]src[._-](r\d+)\.t/i)
```

Based on the file name format, I don't imagine we'll encounter a file name like `musepack_src_vr123.tar.gz`. I'm fine with omitting the`v?` here, as it doesn't use a common version scheme.

Livecheckables/musepack.rb
```diff
@@ -0,0 +1,6 @@
+class Musepack
+  livecheck do
+    url "https://www.musepack.net/index.php?pg=src"
+    regex(/href=.*?musepack[._-]src[._-](r\d+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?musepack(?:[._-]src)?[._-](r\d+)\.t/i)
```

Seeing as the other source files don't include `_src` in the filename, it probably wouldn't be a bad idea to allow matching of a file name like `musepack_r475.tar.gz` as well.

PR #989, opened 2020-06-24
ec2-api-tools: add livecheckable

nandahkrishna:
Adding Livecheckable for `ec2-api-tools`.


nandahkrishna:
While browsing through the website and their documentation, it seems that this is deprecated.

I haven't skipped livecheck here because I was unsure, but the command line tools have a new major version (2) and the documentation link for these tools has a notice saying `(Deprecated)` near it; any thoughts on this @samford?

nandahkrishna:
image
Here's a screenshot and the link to the page is: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Welcome.html

Hope that's helpful; I'll add more sources if I find them.

PR #992, opened 2020-06-24
stress: add livecheckable

nandahkrishna:
Adding Livecheckable for `stress`. By the way, it seems the homepage is down. Don't know if only I'm facing this issue.


samford:
The page (and `/~apw/` user directory) seems to have disappeared somewhere around July/August 2019 and now it gives a "Not Found" error. I'll have to update the homepage to an archived copy unless the site has been moved somewhere else.

PR #996, opened 2020-06-24
node_exporter: add livecheckable

nandahkrishna:
Adding Livecheckable for `node_exporter`.


Livecheckables/node_exporter.rb
```diff
@@ -0,0 +1,6 @@
+class NodeExporter
+  livecheck do
+    url "https://github.com/prometheus/node_exporter.git"
```
samford, line 3:
```suggestion
    url :head
```

PR #998, opened 2020-06-24
httperf: add livecheckable

nandahkrishna:
Adding Livecheckable for `httperf`.

Development has moved from Google Code to https://github.com/httperf/httperf but the problem is that they don't have releases or tags, or any changes.txt we can check for the latest version – that happens to be `0.9.1` which isn't available from Google Code. Have I missed something?

### DIFF
--- a/Livecheckables/httperf.rb
+++ b/Livecheckables/httperf.rb
@@ -1,6 +1,7 @@
 class Httperf
+  # Until the upstream GitHub repository creates a new release (something after
+  # 0.9.0), we're unable to create a check that can identify new versions.
   livecheck do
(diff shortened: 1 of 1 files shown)


Livecheckables/httperf.rb
```diff
@@ -0,0 +1,6 @@
+class Httperf
+  livecheck do
+    url "https://www.googleapis.com/download/storage/v1/b/google-code-archive/o/v2%2Fcode.google.com%2Fhttperf%2Fdownloads-page-1.json\?\&alt\=media"
+    regex(/httperf-v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
  # Until the upstream GitHub repository creates a new release (something after
  # 0.9.0), we're unable to create a check that can identify new versions.
  livecheck do
    skip "No version information available to check"
```

PR #1021, opened 2020-06-26
ttyrec: add livecheckable

nandahkrishna:
Adding Livecheckable for `ttyrec`.

### DIFF
--- a/Livecheckables/ttyrec.rb
+++ b/Livecheckables/ttyrec.rb
@@ -1,6 +1,6 @@
 class Ttyrec
   livecheck do
     url :homepage
-    regex(/href=.*?ttyrec-v?(\d+(?:\.\d+)+)\.t/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/ttyrec.rb
```diff
@@ -0,0 +1,6 @@
+class Ttyrec
+  livecheck do
+    url :homepage
+    regex(/href=.*?ttyrec-v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=['"]?ttyrec-v?(\d+(?:\.\d+)+)\.t/i)
```

There's also a `Term-TtyRec-0.03.tar.gz` archive on the page and we're in the same situation as `ifstat` where the URL is relative and we have to use `href=['"]?` as a leading delimiter.

PR #1010, opened 2020-06-24
Enable multiple strategies

maxim-belkin:
- Allow multiple strategies to match URL
- factor out available strategies into a method
- `strat_matcher` method returns whether specified strategy is applicable to specified url
- warn when no strategies have been matched
- guard `Matched Text on Page` and `Matched Versions` outputs by `Homebrew.args.verbose?`

Formulae that can be used to see the effect of this PR: xmp, wtf, aespipe
 
<details>
<summary><code>brew livecheck -d xmp</code></summary>
<pre><code>
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/xmp.rb
Formula:         xmp
Livecheckable?:  No

URL:             https://git.code.sf.net/p/xmp/xmp-cli.git
\# of strategies: 2
Strategies:      git, sourceforge
Using strategy: git
[...]

PR #1025, opened 2020-06-26
Refactor strategies, add debug and JSON info

samford:
Work on this started months ago when the strategies were all one big if/elsif statement in `heuristic.rb` and the general idea was to refactor the strategies into individual units of some sort (see the tail end of #301).

@maxim-belkin opened a PR (#601) before I had gotten around to creating one for my changes, so I ended up using that PR to close the gap some between the state of livecheck at the time and my refactoring work by moving the strategies into functions in `livecheck_strategy.rb`.

He had suggested that we create individual files for each strategy but I hadn't determined if that was going to be the best approach with the future migration to Homebrew/brew in mind. This ended up being appropriate In the end, after I took some inspiration from `unpack_strategy.rb`.

This creates a `LivecheckStrategy` module and convert the strategies to classes as part of that module, with all of the information relating to a strategy encompassed in its class. This helps to keep things organized while also moving code out of `heuristic.rb`, as everything in that file eventually needs to find a different home (so we can delete it to ease the future migration).

What was once the "heuristic" is now `LivecheckStrategy.from_url()`. The conditions for which a URL matches a strategy is now kept in each strategy's `match?` method. There's also a `LivecheckStrategy.from_symbol()` method, which uses an internal hash that maps symbols to strategies and is necessary for the forthcoming `strategy` addition to the livecheck DSL.

One other thing to note is that strategies now return a hash like `{ :matches => {}, :regex => regex, :url => page_url }` (or `:url => url`). This allows us to print debug output for the URL and regex that may be generated within a strategy, as well as passing it through to the verbose JSON output. Using a hash for this also allows us to extend it in the future, which is necessary for another upcoming feature.

Lastly, this removes the "Matched Text on Page" content from debug output. It was necessary to remove this to keep the debug output coherent, as we can't print the strategy URL and/or regex until after the "Matched Text on Page" output is printed. However, the "Matched Text on Page" output was almost always (if not always) redundant in practice anyway, as the "Matched Versions" were identical, so it was a good idea in its own right.

Overall, this should be easier to maintain and extend. Please let me know if you have any feedback, questions, requests, etc.

### DIFF
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -73,11 +73,13 @@ def latest_version(formula)
     next if original_url.include?("gist.github.com")
 
     url = preprocess_url(original_url)
-    strategy = LivecheckStrategy.from_url(url, livecheck_regex.present?)
+    strategies = LivecheckStrategy.from_url(url, livecheck_regex.present?)
+    strategy = strategies[0]
 
     if Homebrew.args.debug?
       puts "\nURL:              #{original_url}"
       puts "URL (processed):  #{url}" if url != original_url
+      puts "Strategies:       #{strategies.map { |s| s.name.demodulize }}" unless strategies.empty? || !Homebrew.args.verbose?
       puts "Strategy:         #{strategy.nil? ? "None" : strategy::NICE_NAME}"
       puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
     end
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -63,24 +63,28 @@ def latest_version(formula)
   urls ||= checkable_urls(formula)
 
   if Homebrew.args.debug?
-    puts "Formula:          #{formula_name(formula)}"
+    puts "\nFormula:          #{formula_name(formula)}"
     puts "Head only?:       #{!formula.stable?}" unless formula.stable?
     puts "Livecheckable?:   #{has_livecheckable ? "Yes" : "No"}"
   end
 
   urls.each do |original_url|
+    puts "\nURL:              #{original_url}" if Homebrew.args.debug?
+
     # Skip Gists until/unless we create a method of identifying revisions
-    next if original_url.include?("gist.github.com")
+    if original_url.include?("gist.github.com")
+      odebug "Skipping: GitHub Gists are not supported"
+      next
+    end
 
     url = preprocess_url(original_url)
     strategies = LivecheckStrategy.from_url(url, livecheck_regex.present?)
     strategy = strategies[0]
 
     if Homebrew.args.debug?
-      puts "\nURL:              #{original_url}"
       puts "URL (processed):  #{url}" if url != original_url
       puts "Strategies:       #{strategies.map { |s| s.name.demodulize }}" unless strategies.empty? || !Homebrew.args.verbose?
-      puts "Strategy:         #{strategy.nil? ? "None" : strategy::NICE_NAME}"
+      puts "Strategy:         #{strategy.nil? ? "None" : strategy.name.demodulize}"
       puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
     end
 
@@ -107,8 +111,13 @@ def latest_version(formula)
 
     if Homebrew.args.debug? && !match_version_map.empty?
       puts "\nMatched Versions:\n"
-      match_version_map.each do |match, version|
-        puts "#{match} => #{version.inspect}"
+
+      if Homebrew.args.verbose?
+        match_version_map.each do |match, version|
+          puts "#{match} => #{version.inspect}"
+        end
+      else
+        puts match_version_map.values.join(", ")
       end
     end
 
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -15,35 +15,6 @@ GITHUB_SPECIAL_CASES = %w[
   yuicompressor
 ].freeze
 
-GNU_SPECIAL_CASES = %w[
-  kawa
-  lzip
-  numdiff
-  icoutils
-  dvdrtools
-  avrdude
-  oath-toolkit
-  cvs
-  mit-scheme
-  clisp
-].freeze
-
-SOURCEFORGE_SPECIAL_CASES = %w[
-  mikmod
-  log4cpp
-  libwps
-  e2fsprogs
-  potrace
-  remake
-  /avf/
-  /bashdb/
-  /netpbm/
-  opencore-amr
-  liba52.sourceforge.net/
-  foremost.sourceforge.net/
-  bcrypt.sourceforge.net
-].freeze
-
 UNSTABLE_VERSION_KEYWORDS = %w[
   alpha
   beta
@@ -93,9 +64,9 @@ def latest_version(formula)
   urls ||= checkable_urls(formula)
 
   if Homebrew.args.debug?
-    puts "Formula:         #{formula_name(formula)}"
-    puts "Head only?:      #{!formula.stable?}" unless formula.stable?
-    puts "Livecheckable?:  #{has_livecheckable ? "Yes" : "No"}"
+    puts "Formula:          #{formula_name(formula)}"
+    puts "Head only?:       #{!formula.stable?}" unless formula.stable?
+    puts "Livecheckable?:   #{has_livecheckable ? "Yes" : "No"}"
   end
 
   urls.each do |original_url|
@@ -103,42 +74,25 @@ def latest_version(formula)
     next if original_url.include?("gist.github.com")
 
     url = preprocess_url(original_url)
-
-    strategy = if /hackage\.haskell\.org/.match?(url)
-      :hackage_strategy
-    elsif DownloadStrategyDetector.detect(url) <= GitDownloadStrategy
-      :git_strategy
-    elsif /(sourceforge|sf)\.net/.match?(url) &&
-          SOURCEFORGE_SPECIAL_CASES.none? { |sc| url.include? sc }
-      :sourceforge_strategy
-    elsif url =~ /gnu\.org/ && GNU_SPECIAL_CASES.none? { |sc| url.include? sc }
-      :gnu_strategy
-    elsif /files\.pythonhosted\.org/.match?(url)
-      :pypi_strategy
-    elsif /registry\.npmjs\.org/.match?(url)
-      :npm_strategy
-    elsif /download\.gnome\.org/.match?(url)
-      :gnome_strategy
-    elsif /launchpad\.net/.match?(url)
-      :launchpad_strategy
-    elsif %r{www\.apache\.org/dyn}.match?(url)
-      :apache_strategy
-    elsif %r{bitbucket\.org(/[^/]+){4}\.\w+}.match?(url)
-      :bitbucket_strategy
-    elsif livecheck_regex
-      :page_match_strategy
-    end
+    strategy = LivecheckStrategy.from_url(url, !livecheck_regex.nil?)
 
     if Homebrew.args.debug?
-      puts "\nURL:             #{original_url}"
-      puts "URL (processed): #{url}" if url != original_url
-      puts "Strategy:        #{strategy.nil? ? "None" : strategy.to_s.delete_suffix("_strategy")}"
-      puts "Regex:           #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
+      puts "\nURL:              #{original_url}"
+      puts "URL (processed):  #{url}" if url != original_url
+      puts "Strategy:         #{strategy.nil? ? "None" : strategy::NICE_NAME}"
+      puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
     end
 
     next if strategy.nil?
 
-    match_version_map = Symbol.send(strategy, url, livecheck_regex)
+    strategy_data = strategy.find_versions(url, livecheck_regex)
+    match_version_map = strategy_data[:matches]
+    regex = strategy_data[:regex]
+
+    if Homebrew.args.debug?
+      puts "URL (strategy):   #{strategy_data[:url]}" if strategy_data[:url] != url
+      puts "Regex (strategy): #{strategy_data[:regex].inspect}\n" if strategy_data[:regex] != livecheck_regex
+    end
 
     empty_version = Version.new("")
     match_version_map.delete_if do |_match, version|
@@ -168,10 +122,11 @@ def latest_version(formula)
         "url"      => {
           "original" => original_url,
         },
-        "strategy" => strategy.nil? ? nil : strategy.to_s.delete_suffix("_strategy"),
+        "strategy" => strategy.nil? ? nil : strategy::NAME,
       }
       version_info["meta"]["url"]["processed"] = url if url != original_url
-      version_info["meta"]["regex"] = livecheck_regex.inspect unless livecheck_regex.nil?
+      version_info["meta"]["url"]["strategy"] = strategy_data[:url] if strategy_data[:url] != url
+      version_info["meta"]["regex"] = regex.inspect unless regex.nil?
     end
 
     return version_info
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -132,10 +132,11 @@ def latest_version(formula)
         "url"      => {
           "original" => original_url,
         },
-        "strategy" => strategy.nil? ? nil : strategy::NAME,
+        "strategy" => strategy.nil? ? nil : strategy.name.demodulize,
       }
       version_info["meta"]["url"]["processed"] = url if url != original_url
       version_info["meta"]["url"]["strategy"] = strategy_data[:url] if strategy_data[:url] != url
+      version_info["meta"]["strategies"] = strategies.map { |s| s.name.demodulize } unless strategies.empty?
       version_info["meta"]["regex"] = regex.inspect unless regex.nil?
     end
 
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -1,4 +1,3 @@
-require_relative "livecheck_strategy"
 require "utils"
 
 GITHUB_SPECIAL_CASES = %w[
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -74,7 +74,7 @@ def latest_version(formula)
     next if original_url.include?("gist.github.com")
 
     url = preprocess_url(original_url)
-    strategy = LivecheckStrategy.from_url(url, !livecheck_regex.nil?)
+    strategy = LivecheckStrategy.from_url(url, livecheck_regex.present?)
 
     if Homebrew.args.debug?
       puts "\nURL:              #{original_url}"
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -83,7 +83,9 @@ def latest_version(formula)
 
     if Homebrew.args.debug?
       puts "URL (processed):  #{url}" if url != original_url
-      puts "Strategies:       #{strategies.map { |s| s.name.demodulize }}" unless strategies.empty? || !Homebrew.args.verbose?
+      unless strategies.empty? || !Homebrew.args.verbose?
+        puts "Strategies:       #{strategies.map { |s| s.name.demodulize }.join(", ")}"
+      end
       puts "Strategy:         #{strategy.nil? ? "None" : strategy.name.demodulize}"
       puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
     end
--- a/livecheck/heuristic.rb
+++ b/livecheck/heuristic.rb
@@ -84,7 +84,7 @@ def latest_version(formula)
     if Homebrew.args.debug?
       puts "URL (processed):  #{url}" if url != original_url
       unless strategies.empty? || !Homebrew.args.verbose?
-        puts "Strategies:       #{strategies.map { |s| s.name.demodulize }.join(", ")}"
+        puts "Strategies:       #{strategies.map { |s| s::NAME }.join(", ")}"
       end
       puts "Strategy:         #{strategy.nil? ? "None" : strategy.name.demodulize}"
       puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
@@ -134,11 +134,11 @@ def latest_version(formula)
         "url"      => {
           "original" => original_url,
         },
-        "strategy" => strategy.nil? ? nil : strategy.name.demodulize,
+        "strategy" => strategy.nil? ? nil : strategy::NAME,
       }
       version_info["meta"]["url"]["processed"] = url if url != original_url
       version_info["meta"]["url"]["strategy"] = strategy_data[:url] if strategy_data[:url] != url
-      version_info["meta"]["strategies"] = strategies.map { |s| s.name.demodulize } unless strategies.empty?
+      version_info["meta"]["strategies"] = strategies.map { |s| s::NAME } unless strategies.empty?
       version_info["meta"]["regex"] = regex.inspect unless regex.nil?
     end
 
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -25,9 +25,11 @@ module LivecheckStrategy
 
     usable_strategies << strategies[:page_match] if strategies.key?(:page_match) && regex_provided
 
+    # Sort usable strategies in descending order by priority, using 5 as the
+    # default when a PRIORITY isn't provided in the LivecheckStrategy itself.
     usable_strategies.sort_by do |strategy|
-      (strategy.const_defined?(:PRIORITY) ? strategy::PRIORITY : 5)
-    end.reverse
+      (strategy.const_defined?(:PRIORITY) ? -strategy::PRIORITY : -5)
+    end
   end
 end
 
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -10,7 +10,7 @@ module LivecheckStrategy
       strategy = LivecheckStrategy.const_get(strategy_symbol)
       @strategies[key] = strategy
     end
-    @strategies.freeze
+    @strategies
   end
   private_class_method :strategies
 
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -32,9 +32,3 @@ module LivecheckStrategy
 end
 
 Dir.glob(File.join(__dir__, "livecheck_strategy", "*.rb"), &method(:require))
-
-if Homebrew.args.tap
-  current_tap = Tap.fetch(Homebrew.args.tap)
-  tap_strategy_path = File.join(current_tap.path, "livecheck_strategy")
-  Dir.glob(File.join(tap_strategy_path, "*.rb"), &method(:require)) if Dir.exist?(tap_strategy_path)
-end
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -24,30 +24,28 @@ module LivecheckStrategy
 
   def self.from_url(url, regex_provided = nil)
     if Hackage.match?(url)
-      return Hackage
+      Hackage
     elsif Git.match?(url)
-      return Git
+      Git
     elsif SourceForge.match?(url)
-      return SourceForge
+      SourceForge
     elsif Gnu.match?(url)
-      return Gnu
+      Gnu
     elsif PyPI.match?(url)
-      return PyPI
+      PyPI
     elsif Npm.match?(url)
-      return Npm
+      Npm
     elsif Gnome.match?(url)
-      return Gnome
+      Gnome
     elsif Launchpad.match?(url)
-      return Launchpad
+      Launchpad
     elsif Apache.match?(url)
-      return Apache
+      Apache
     elsif Bitbucket.match?(url)
-      return Bitbucket
+      Bitbucket
     elsif regex_provided
-      return PageMatch
+      PageMatch
     end
-
-    nil
   end
 end
 
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -25,11 +25,9 @@ module LivecheckStrategy
 
     usable_strategies << strategies[:page_match] if strategies.key?(:page_match) && regex_provided
 
-    usable_strategies.sort_by! do |strategy|
+    usable_strategies.sort_by do |strategy|
       (strategy.const_defined?(:PRIORITY) ? strategy::PRIORITY : 5)
-    end.reverse!
-
-    usable_strategies[0]
+    end.reverse
   end
 end
 
--- a/livecheck/livecheck_strategy.rb
+++ b/livecheck/livecheck_strategy.rb
@@ -1,218 +1,64 @@
-# Formulae that do not use GNOME's "even-numbered minor is stable" scheme
-GNOME_DEVEL_ALLOWLIST = [
-  "gcab",
-  "gtk-doc",
-  "gtk-mac-integration",
-  "libart_lgpl", # The package name for libart is "libart_lgpl"
-  "libepoxy",
-].freeze
+# frozen_string_literal: true
 
-def apache_strategy(url, regex = nil)
-  match_version_map = {}
+module LivecheckStrategy
+  def self.strategies
+    @strategies ||= {
+      :apache      => Apache,
+      :bitbucket   => Bitbucket,
+      :git         => Git,
+      :gnome       => Gnome,
+      :gnu         => Gnu,
+      :hackage     => Hackage,
+      :launchpad   => Launchpad,
+      :npm         => Npm,
+      :page_match  => PageMatch,
+      :pypi        => PyPI,
+      :sourceforge => SourceForge,
+    }.freeze
+  end
+  private_class_method :strategies
 
-  path, prefix, suffix = url.match(%r{path=(.+?)/([^/]*?)\d+(?:\.\d+)+(/|[^/]*)})[1, 3]
-  page_url = "https://archive.apache.org/dist/#{path}/"
-
-  regex ||= /href="#{Regexp.escape(prefix)}(\d+(?:\.\d+)+)#{Regexp.escape(suffix)}/
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
+  def self.from_symbol(symbol)
+    strategies[symbol]
   end
 
-  match_version_map
-end
-
-def bitbucket_strategy(url, regex = nil)
-  match_version_map = {}
-
-  path, kind, suffix =
-    url.match(%r{bitbucket\.org/(.+?)/(get|downloads)/(?:.*?[-_])?v?\d+(?:\.\d+)+([^/]+)})[1, 3]
-  page_url = "https://bitbucket.org/#{path}/downloads/"
-  page_url << "?tab=tags" if kind == "get"
-
-  regex ||= /(\d+(?:\.\d+)+)#{Regexp.escape(suffix)}"/
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def git_strategy(url, regex = nil)
-  match_version_map = {}
-
-  tags = git_tags(url, regex)
-  tags_only_debian = tags.all? { |tag| tag.start_with?("debian/") }
-
-  tags.each do |tag|
-    # Move to the next one if tag actually is prefixed with 'debian/'
-    # and upstream does not do only 'debian/' prefixed tags
-    next if tag =~ %r{debian/} && !tags_only_debian
-
-    captures = tag.scan(regex) if regex
-    tag_cleaned = if captures &&
-                     !captures.empty? &&
-                     captures[0].is_a?(Array)
-      # Use the first capture group as the version
-      captures[0][0]
-    else
-      # Remove any character before the first number
-      tag[/\D*(.*)/, 1]
+  def self.from_url(url, regex_provided = nil)
+    if Hackage.match?(url)
+      return Hackage
+    elsif Git.match?(url)
+      return Git
+    elsif SourceForge.match?(url)
+      return SourceForge
+    elsif Gnu.match?(url)
+      return Gnu
+    elsif PyPI.match?(url)
+      return PyPI
+    elsif Npm.match?(url)
+      return Npm
+    elsif Gnome.match?(url)
+      return Gnome
+    elsif Launchpad.match?(url)
+      return Launchpad
+    elsif Apache.match?(url)
+      return Apache
+    elsif Bitbucket.match?(url)
+      return Bitbucket
+    elsif regex_provided
+      return PageMatch
     end
 
-    match_version_map[tag] = Version.new(tag_cleaned)
-  rescue TypeError
     nil
   end
-
-  match_version_map
 end
 
-def gnome_strategy(url, regex = nil)
-  match_version_map = {}
-
-  package = url.match(%r{/sources/(.*?)/})[1]
-  page_url = "https://download.gnome.org/sources/#{package}/cache.json"
-
-  # Restrict versions to even numbered minor versions (except x.90+)
-  regex ||= if GNOME_DEVEL_ALLOWLIST.include?(package)
-    /#{Regexp.escape(package)}-(\d+(?:\.\d+)+)\.t/
-  else
-    /#{Regexp.escape(package)}-(\d+\.([0-8]\d*?)?[02468](?:\.\d+)*?)\.t/
-  end
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def gnu_strategy(url, regex = nil)
-  match_version_map = {}
-
-  project_name_regexps = [
-    %r{/(?:software|gnu)/(.*?)/},
-    %r{//(.*?)\.gnu\.org(?:/)?$},
-  ]
-  match_list = project_name_regexps.map do |r|
-    url.match(r)
-  end.compact
-
-  puts "Multiple project names found: #{match_list}" if match_list.length > 1 && Homebrew.args.debug?
-
-  unless match_list.empty?
-    project_name = match_list[0][1]
-    page_url = "http://ftp.gnu.org/gnu/#{project_name}/?C=M&O=D"
-
-    regex ||= /#{project_name}-(\d+(?:\.\d+)*)/
-
-    page_matches(page_url, regex).each do |match|
-      version = Version.new(match)
-      match_version_map[match] = version
-    end
-  end
-
-  match_version_map
-end
-
-def hackage_strategy(url, _regex = nil)
-  match_version_map = {}
-
-  package = ((url.split("/")[4]).split("-")[0..-2]).join("-")
-  page_url = "https://hackage.haskell.org/package/#{package}/src"
-
-  regex ||= %r{<h3>#{package}-(.*?)/?</h3>}i
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def launchpad_strategy(url, regex = nil)
-  match_version_map = {}
-
-  package = url.match(%r{launchpad\.net/([^/]*)})[1]
-  page_url = "https://launchpad.net/#{package}"
-
-  regex ||= %r{<div class="version">\s*Latest version is (.+)\s*</div>}
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def npm_strategy(url, regex = nil)
-  match_version_map = {}
-
-  package = url.split("/")[3..-3].reject { |s| s == "-" }.join("/")
-  page_url = "https://www.npmjs.com/package/#{package}?activeTab=versions"
-
-  regex ||= %r{/package/#{package}/v/(\d+(?:\.\d+)+)"}
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def page_match_strategy(url, regex = nil)
-  match_version_map = {}
-
-  page_matches(url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def pypi_strategy(url, regex = nil)
-  match_version_map = {}
-
-  package = url[%r{https://files.pythonhosted.org/packages/.*/.*/(.*)-.*}, 1]
-  page_url = "https://pypi.org/project/#{package}"
-
-  regex ||= /#{package} ([0-9.]+)/
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
-
-def sourceforge_strategy(url, regex = nil)
-  match_version_map = {}
-
-  project_name = if url.include?("/project")
-    url.match(%r{/projects?/([^/]+)/})[1]
-  elsif url.include?(".net/p/")
-    url.match(%r{\.net/p/([^/]+)/})[1]
-  else
-    url.match(%r{\.net(?::/cvsroot)?/([^/]+)})[1]
-  end
-  page_url = "https://sourceforge.net/projects/#{project_name}/rss"
-
-  regex ||= %r{url=.+?/#{project_name}/files/.*?[-_/](\d+(?:[-.]\d+)+)[-_/%.]}i
-
-  page_matches(page_url, regex).each do |match|
-    version = Version.new(match)
-    match_version_map[match] = version
-  end
-
-  match_version_map
-end
+require_relative "livecheck_strategy/apache.rb"
+require_relative "livecheck_strategy/bitbucket.rb"
+require_relative "livecheck_strategy/git.rb"
+require_relative "livecheck_strategy/gnome.rb"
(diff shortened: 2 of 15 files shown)


livecheck/livecheck_strategy/git.rb
```diff
@@ -0,0 +1,39 @@
...
+        next if tag =~ %r{debian/} && !tags_only_debian
+
+        captures = tag.scan(regex) if regex
+        tag_cleaned = if captures && !captures.empty? && captures[0].is_a?(Array)
+          # Use the first capture group as the version
+          captures[0][0]
+        else
+          # Remove any character before the first number
+          tag[/\D*(.*)/, 1]
+        end
```
maxim-belkin, line 29: First comment "Use the first capture group as the **version**" contradicts variable name `tag_cleaned`

```suggestion
        captures = regex.present? ? tag.scan(regex) : []
        tag_cleaned = if captures[0].is_a?(Array)
          captures[0][0]  # Extract the first capture group
        else
          tag[/\D*(.*)/, 1]  # Remove non-digits from the beginning of the tag
        end
```

livecheck/livecheck_strategy/gnome.rb
```diff
@@ -0,0 +1,41 @@
...
+    NAME = NICE_NAME.downcase
+
+    # Formulae that do not use GNOME's "even-numbered minor is stable" scheme
+    GNOME_DEVEL_ALLOWLIST = [
+      "gcab",
+      "gtk-doc",
+      "gtk-mac-integration",
+      "libart_lgpl", # The package name for libart is "libart_lgpl"
+      "libepoxy",
+    ].freeze
```
maxim-belkin, line 15:
```suggestion
    # Formulae that do not use GNOME's "even-numbered minor is stable" scheme
    # "libart_lgpl" is the package name for libart
    GNOME_DEVEL_ALLOWLIST = %w[
      gcab
      gtk-doc
      gtk-mac-integration
      libart_lgpl
      libepoxy
    ].freeze
```
```

File: livecheck/livecheck_strategy.rb [hunk 3]
```diff
@@ -1,218 +1,64 @@
...
+    elsif Gnome.match?(url)
+      return Gnome
+    elsif Launchpad.match?(url)
+      return Launchpad
+    elsif Apache.match?(url)
+      return Apache
+    elsif Bitbucket.match?(url)
+      return Bitbucket
+    elsif regex_provided
+      return PageMatch
```
maxim-belkin, line 47: You can remove all "return"s here and move `nil` to under final "else"

```
    if Hackage.match?(url)
      Hackage
    elsif Git.match?(url)
      Git
    elsif SourceForge.match?(url)
      SourceForge
    elsif Gnu.match?(url)
      Gnu
    elsif PyPI.match?(url)
      PyPI
    elsif Npm.match?(url)
      Npm
    elsif Gnome.match?(url)
      Gnome
    elsif Launchpad.match?(url)
      Launchpad
    elsif Apache.match?(url)
      Apache
    elsif Bitbucket.match?(url)
      Bitbucket
    elsif regex_provided
      PageMatch
    else
      nil
    end
```

livecheck/livecheck_strategy/gnome.rb
```diff
@@ -0,0 +1,41 @@
...
+        /#{Regexp.escape(package)}-(\d+\.([0-8]\d*?)?[02468](?:\.\d+)*?)\.t/
+      end
+
+      match_data = { :matches => {}, :regex => regex, :url => page_url }
+      page_matches(page_url, regex).each do |match|
+        version = Version.new(match)
+        match_data[:matches][match] = version
+      end
+
+      match_data
```
maxim-belkin, line 38: This final piece repeats across all files under `livecheck/livecheck_strategy` and is, in fact what `page_match` "strategy" does. Can you factor it out so that creation of new strategies does not have to repeat this mechanical step?

livecheck/livecheck_strategy/sourceforge.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module LivecheckStrategy
+  class SourceForge
+    NICE_NAME = "SourceForge"
+    NAME = NICE_NAME.downcase
+
+    SOURCEFORGE_SPECIAL_CASES = %w[
```
maxim-belkin, line 8: Does it make sense to use unique names for special cases when these are in separate files?
Would "SPECIAL_CASES" be sufficient?

livecheck/heuristic.rb
```diff
@@ -93,52 +64,35 @@ def latest_version(formula)
...
-    elsif /launchpad\.net/.match?(url)
-      :launchpad_strategy
-    elsif %r{www\.apache\.org/dyn}.match?(url)
-      :apache_strategy
-    elsif %r{bitbucket\.org(/[^/]+){4}\.\w+}.match?(url)
-      :bitbucket_strategy
-    elsif livecheck_regex
-      :page_match_strategy
-    end
+    strategy = LivecheckStrategy.from_url(url, !livecheck_regex.nil?)
```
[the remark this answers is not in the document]
samford: Ah, this is one of those cases where my system Ruby (2.7.1p83) differs from Homebrew's Ruby. Within `irb`, `nil.present?` gives `NoMethodError (undefined method `present?' for nil:NilClass)`. Within `brew irb`, the same gives `false`.

This is also true for `nil.blank?`, which is why I generally avoid using it unless I've already checked the object's class (`livecheck_url.is_a?(String) && !livecheck_url.blank?`) or I'm certain the value can't be `nil` (e.g., due to object initialization).

I think it's better to rely on the standard Ruby behavior instead of the Homebrew-specific one, so we're not caught out in the future. What are your thoughts on this?

livecheck/heuristic.rb
```diff
@@ -168,10 +122,11 @@ def latest_version(formula)
         "url"      => {
           "original" => original_url,
         },
-        "strategy" => strategy.nil? ? nil : strategy.to_s.delete_suffix("_strategy"),
+        "strategy" => strategy.nil? ? nil : strategy::NAME,
       }
       version_info["meta"]["url"]["processed"] = url if url != original_url
-      version_info["meta"]["regex"] = livecheck_regex.inspect unless livecheck_regex.nil?
+      version_info["meta"]["url"]["strategy"] = strategy_data[:url] if strategy_data[:url] != url
+      version_info["meta"]["regex"] = regex.inspect unless regex.nil?
```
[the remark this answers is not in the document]
samford: My thinking is that the information in the default JSON output mirrors the information in the default livecheck output. This covers livecheck's basic behavior (comparing the Formula version to a remote version) and it's sufficient for anyone with those basic needs.

When a user needs additional information, they use the `--debug` flag. In a similar fashion, when a user working with the JSON output needs the same additional information, they use `--json --verbose`. The way I've been approaching it, `--verbose` for JSON is the same as `--debug` for normal output. [I'll be explaining this in documentation in the future, so users don't have to dig into the code.]

My other thought is that the verbose output is regularly expanding to meet additional needs, so users with basic needs shouldn't have to deal with the added burden now or in the future. A full livecheck run of the default JSON output is around 694 KB whereas a full run of the verbose JSON output is around 1.2 MB currently and will continue to expand in the future.

I think there's value in limiting the default JSON output but I'm open to revisiting this if enough users feel otherwise. I just feel that doubling the size of the output (as will happen after some forthcoming additions) is enough to warrant the current approach and using `brew livecheck -jv` isn't a burden.

livecheck/livecheck_strategy/git.rb
```diff
@@ -0,0 +1,39 @@
...
+    NICE_NAME = "Git"
+    NAME = NICE_NAME.downcase
+
+    def self.match?(url)
+      DownloadStrategyDetector.detect(url) <= GitDownloadStrategy
+    end
+
+    def self.find_versions(url, regex)
+      tags = git_tags(url, regex)
+      tags_only_debian = tags.all? { |tag| tag.start_with?("debian/") }
```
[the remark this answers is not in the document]
samford: Good idea! I may prefer a name like `has_non_debian_tag` or `found_non_debian_tag`, though.

I don't want to make too many changes to the strategies in this PR and I feel like this is starting to get outside of the scope of this PR. However, I would like to see this change merged, so please make a follow-up PR for it after this one is merged.

livecheck/livecheck_strategy/git.rb
samford: I agree that the language in the comment put the cart before the horse (as creating the version from `tag_cleaned` doesn't happen until after). "Extract the first capture group" makes better sense to me as well.

I would have thought that `captures[0].is_a?(Array)` would have given an error when `captures` is empty but it surprisingly doesn't. Good to know.

livecheck/livecheck_strategy.rb
samford: Whoops, I've been looking at all of this for far too long to have made such a simple mistake, ahah. I've very thankful for the extra set of eyes.

With this format, we can omit the `else` `nil` altogether (as `nil` will be returned by default when no conditions match).

livecheck/livecheck_strategy/sourceforge.rb
samford: Yes, good catch! The `SOURCEFORGE_SPECIAL_CASES` name was simply a holdover from when the array was separate from the strategy and it definitely makes sense to drop the leading name in this new context (e.g., `SPECIAL_CASES`).

I will also apply this to `GNOME_DEVEL_ALLOWLIST`, `GNU_PROJECT_NAME_REGEXES`, and `GNU_SPECIAL_CASES`.

livecheck/heuristic.rb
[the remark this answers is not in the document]
maxim-belkin: I was asked to change `!...nil?` to `.present?` a few times myself so I expect the same requests to surface up here when Livecheck is merged into Homebrew/brew.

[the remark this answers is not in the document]
samford: I've raised the question to the other maintainers to see what the prevailing view is and I'll update this accordingly. Thanks for bringing it to my attention, as we're certainly trying to do all we can to ease the migration to Homebrew/brew.

livecheck/heuristic.rb
```diff
@@ -93,52 +64,35 @@ def latest_version(formula)
   urls ||= checkable_urls(formula)
 
   if Homebrew.args.debug?
-    puts "Formula:         #{formula_name(formula)}"
-    puts "Head only?:      #{!formula.stable?}" unless formula.stable?
-    puts "Livecheckable?:  #{has_livecheckable ? "Yes" : "No"}"
+    puts "Formula:          #{formula_name(formula)}"
+    puts "Head only?:       #{!formula.stable?}" unless formula.stable?
+    puts "Livecheckable?:   #{has_livecheckable ? "Yes" : "No"}"
```
maxim-belkin, line 69: Are these extra spaces necessary?

livecheck/livecheck_strategy/gnome.rb
samford: This is definitely something that's been on my mind and I wanted to work on addressing it but I didn't want to further delay creating this PR by working on it.

All of the uses of this snippet are standardized, so we can currently replace them with `PageMatch.find_versions(page_url, regex)`, which would be functionally identical. I've tested this and it seems to work but I want to spend a little more time considering the implications.

Just off the top of my head, my main concern would be how we would handle adding additional information to the `match_data` hash but, thinking about it, it's as simple as:

```ruby
match_data = PageMatch.find_versions(page_url, regex)
match_data[:something_extra] = "something"
```

Seems pretty straightforward but I'm going to think on it just a little bit more before putting the changes in place. Thanks for the reminder.

livecheck/heuristic.rb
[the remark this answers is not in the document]
maxim-belkin: > I personally think there's value in limiting the default JSON output

I agree with that. What I'm saying is that output fields in APIs (and I think JSON is a step toward an API approach) should be controlled via verbose/debug switches and not depend on whether regex is present or not.

livecheck/heuristic.rb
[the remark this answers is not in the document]
samford: It was brought to my attention that Homebrew uses `activesupport`, which certainly explains the behavioral difference. Considering the existing usage within Homebrew that assumes the `activesupport` behavior, I'll fall in line going forward.

livecheck/livecheck_strategy/gnu.rb
```diff
@@ -0,0 +1,57 @@
...
+    def self.match?(url)
+      url =~ /gnu\.org/ && SPECIAL_CASES.none? { |sc| url.include? sc }
+    end
+
+    def self.find_versions(url, regex)
+      match_data = { :matches => {}, :regex => regex, :url => url }
+
+      match_list = PROJECT_NAME_REGEXES.map do |r|
+        url.match(r)
+      end.compact
```
maxim-belkin, line 37:
```suggestion
      match_list = PROJECT_NAME_REGEXES.map { |r| url.match(r) }.compact
```

livecheck/livecheck_strategy/gnome.rb
maxim-belkin: In a generic case, these `find_versions` functions:

1. set up `page_url`
2. set up `regex`
3. call that common `PageMatch.find_versions(page_url, regex)` piece.

There are exceptions (Git, Gnu) but if we can bring all strategies to using the above 3 procedures, we could then do something like this:

```ruby
def self.find_versions(url, regex)
    page_url, regex = self.get_page_url_regex(url, regex)
    match_data = PageMatch.find_versions(page_url, regex)
end

This piece would be the same for all strategies and we can factor it even further.
```

livecheck/heuristic.rb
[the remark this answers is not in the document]
samford: Ah, I think I'm beginning to understand your meaning. It sounds like you're suggesting that it may be more appropriate to always include all of the possible keys (e.g., `version_info["meta"]["regex"]`) in the verbose JSON output and simply have them be `nil`/`null` when there's no value.

If the output is sometimes `nil`/`null`, the consumer would seemingly need to ensure the value is a string before using it, so this doesn't strike me as much different than requiring the consumer to check for the existence of an optional key before using it. In the latter case, if the key exists, then the value is a string.

Unless my thinking here is inaccurate, it seems like both approaches involve checks of some sort. If that's true, then optional inclusion (as happens here) would seem to be more beneficial because it minimizes the size of the JSON output. Including the additional data just for a `null` value will increase the size of the output and I'm not sure if that's a good tradeoff.

I'm open to the idea and I feel that this an important discussion, so I'll create an issue where we can continue talking about this once this PR is done. If you have some time to explore some high-profile JSON APIs and compare whether they use a `null` vs. omitted approach, that would be helpful (if not, I'll try to look into it when I have the time).

Beyond this, once things settle a bit, I had intended to provide documentation that shows the entire verbose output and I was planning to mark optional values, so that consumers can know what they could rely on and what may not be present.

livecheck/livecheck_strategy/gnome.rb
samford: Hmm, yeah, I can certainly see that there may be some value in further decomposing the relevant `find_versions` methods, to make them more understandable through standardization. Let me tinker with this a bit.

livecheck/heuristic.rb
```diff
@@ -93,52 +64,35 @@ def latest_version(formula)
...
   end
 
   urls.each do |original_url|
     # Skip Gists until/unless we create a method of identifying revisions
     next if original_url.include?("gist.github.com")
```
maxim-belkin, line 74: Shall we add a message her, e.g. `odebug "Skipping this URL because it matches 'gist.github.com'"?

livecheck/heuristic.rb
```diff
@@ -93,52 +64,35 @@ def latest_version(formula)
...
-      puts "URL (processed): #{url}" if url != original_url
-      puts "Strategy:        #{strategy.nil? ? "None" : strategy.to_s.delete_suffix("_strategy")}"
-      puts "Regex:           #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
+      puts "\nURL:              #{original_url}"
+      puts "URL (processed):  #{url}" if url != original_url
+      puts "Strategy:         #{strategy.nil? ? "None" : strategy::NICE_NAME}"
+      puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
     end
 
     next if strategy.nil?
```
maxim-belkin, line 86: I feel like we have to add a warning ("No strategy found") here.

livecheck/heuristic.rb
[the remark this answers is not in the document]
maxim-belkin: I see some "null"s at https://api.github.com/orgs/Homebrew and https://api.github.com/orgs/Homebrew/events
But what we do depends on whether things like `regex`, `url`, and `strategy_data[:url]` should be part of our standard, verbose, or debug output. But we can discuss this later. It's not critical.

livecheck/livecheck_strategy/gnome.rb
samford: I think the resulting `find_versions()` methods with the use of `PageMatch.find_versions(page_url, regex)` is understandable enough (see the latest commits) and I worry that factoring out the `page_url` and `regex` creation into a separate method may be going a bit too far at this point.

I could maybe see revisiting this in the future if this type of `find_versions` method becomes more complex across the board (something more involved than `Gnu`) but I don't feel like this would be buying us much right now and may be a bit overkill.

That said, this is just my view and it's subjective but I feel like we have to draw a line somewhere, otherwise we may be over-optimizing. If others review this and support factoring out `page_url` and `regex` creation, I'll make the change.

livecheck/livecheck_strategy/sourceforge.rb
```diff
@@ -0,0 +1,45 @@
...
+    ].freeze
+    private_constant :SPECIAL_CASES
+
+    def self.match?(url)
+      /(sourceforge|sf)\.net/.match?(url) &&
+        SPECIAL_CASES.none? { |sc| url.include? sc }
+    end
+
+    def self.find_versions(url, regex)
+      project_name = if url.include?("/project")
```
maxim-belkin, line 31: Shall we call it `package` for consistency with other strategies?

samford: This is the SourceForge project name, so I feel the current name is semantically-accurate and provides better clarity of what the code is doing. This isn't a package [name], so I don't think it would be appropriate to name it `package` for the sake of standardization.

livecheck/heuristic.rb
samford: When no strategy applies to a URL, the output looks like this (using `wmctrl` as an example):

```
Formula:          wmctrl
Livecheckable?:   No

URL:              https://sites.google.com/site/tstyblo/wmctrl/wmctrl-1.07.tar.gz
Strategy:         None

URL:              https://sites.google.com/site/tstyblo/wmctrl
Strategy:         None
Error: wmctrl: Unable to get versions
```

I think this makes it clear that no strategy is being found for these URLs (i.e., it's not getting lost in other debug output) and the ending `Error: wmctrl: Unable to get versions` error is appropriate to me.

I don't think something like this would be any more useful than the above:

```
URL:              https://sites.google.com/site/tstyblo/wmctrl/wmctrl-1.07.tar.gz
Strategy:         None
No strategy found
```

When there's no livecheckable, it's common for livecheck to skip URLs that don't match a strategy as part of trying the available URLs. I don't feel that URLs without a matching strategy are something that we should be warning about.

Have I understood your intention here or were you thinking something different?

livecheck/livecheck_strategy/sourceforge.rb
maxim-belkin: This is up to you but in my view naming decisions made by SourceForge should not play a role here because this is not a project nor a package but a string and we call `package`. I see some use in this being factored out into `self.package` but it looks like you disagree and I'm OK with that.

livecheck/heuristic.rb
samford: Though this currently doesn't impact any formulae, I can get behind the idea of making this apparent in the debug output somehow (for when it does apply to a formula).

Let me tinker with this a bit to figure out how to approach this. Skipping `gist.github.com` URLs was a stopgap measure and I may be able to replace it with something better.

At worst, I'll move the debug output that prints the URL above the `next` statement and add an `odebug` message.

livecheck/heuristic.rb
maxim-belkin: You've convinced me. 👍

maxim-belkin:
```ruby
next true if version == empty_version
```
can be changed to 
```ruby
next if version.to_s.empty?
```
and we can then remove `empty_version = Version.new("")`

livecheck/livecheck_strategy/sourceforge.rb
samford: The `package` variables should really be named `package_name` but that's not the point you're making.

It's possible to standardize on a generic name like `upstream_name` for these variables but I have to wonder if this is improves or worsens the clarity of the code compared to `project_name` and `package_name`. I'm not convinced that it's worth it and I feel like we're maybe getting out in the weeds here.

I do feel that creating something like a `self.upstream_name(url)` method would be premature optimization at this point. Until we have a need to use the upstream name elsewhere and/or factoring out this code would noticeably improve clarity/understanding, I don't think it's something we should pursue (especially since some of these are just one-liners).

maxim-belkin: OK, let's leave it as is. This is indeed a very minor issue.

samford: I did update the `package` variable names to `package_name`, as it's a name rather than a package. This brings it in line with `project_name` and may also improve clarity a bit.

maxim-belkin:
We need to include some logic to allow taps define their own strategies.

livecheck/livecheck_strategy/sourceforge.rb
maxim-belkin: thank you. It does improve clarity.

samford:
> We need to include some logic to allow taps define their own strategies.

Just from a cursory glance, I imagine this would take some notable work to enable. Even if it doesn't, I feel that it would make more sense as a follow-up PR after this is merged, as it may be beyond the scope of this one.

I would welcome a PR for this sometime after this is merged but I'm not likely to work on this myself in the immediate future. There are a number of more important changes that I would like to get done, so investing my time in that is more valuable at the present moment.

samford:
> ```ruby
> next true if version == empty_version
> ``` 
> 
> can be changed to
>
> ```ruby
> next true if version.to_s.empty?
> ```
> 
> and we can then remove `empty_version = Version.new("")`

I created a PR (Homebrew/brew#7846) to add `blank?` and/or `empty?` methods to `Version`. There's no guarantee that this will get merged but it would allow us to do something like `next true if version.empty?` instead and avoid the extra string allocation that `version.to_s.empty?` does.

maxim-belkin:
> Just from a cursory glance, I imagine this would take some notable work to enable. 

Per-tap strategies are important because Homebrew is more than just (brew + homebrew-core). This feature shouldn't require a lot of code similarly to how per-tap livecheckables (21b6e5121168a9646d461eac5781bfa316b5aec4) weren't that difficult to implement but provided a lot of benefits for taps. Probably an additional require in `cmd/livecheck.rb` guarded by `if Homebrew.args.tap`.

> What's primarily standing in the way of this right now is the hard-coded strategies in `livecheck_strategy.rb`.

Yes. What's also blocking it is that strategy determination is implemented as an if/else block instead of `delete_if/keep_if` (and then returning just a single value for one-strategy approach or the entire array of suitable strategies for multi-strategy approach). If we could populate the pool of strategies and then use that pool in `from_url`, then tap could append / prepend to that pool and we'd be done.

> I created a PR ([Homebrew/brew#7846](https://github.com/Homebrew/brew/pull/7846)) to add `blank?` and/or `empty?` methods to `Version`. There's no guarantee that this will get merged but it would allow us to do something like `next true if version.empty?` instead and avoid the extra string allocation that `version.to_s.empty?` does.

I actually thought about that but then decided that it's too much work :)

samford:
Just to give you an update, I've done most of the work in the latest commit to enable taps to provide strategies. At the moment, all a tap would need to do is have a `livecheck_strategy` folder with strategies in them and they'll automatically load. Does this seem logical?

[I'm still using `require_relative` statements for the built-in strategies here, until I can confirm whether the overly-clever way of requiring strategies (without being explicit) will work for documentation purposes.]

The tricky part was that `Homebrew.args.tap` wasn't available to `livecheck_strategy.rb`, since args aren't parsed until the livecheck command is executed (i.e., it's not available when `livecheck_strategy.rb` is required). I worked around this by requiring the file after `livecheck_args.parse`, as I couldn't find another working solution. It's not quite ideal but it works for now.

---

That said, though the strategies are loaded and appear in `strategies`, they're not yet usable, as I still need to rework `#from_url`. I've already written an alternative implementation of this method that uses `strategies` (instead of the hardcoded if/else block) but I need to do some research before we can commit to using it.

I'm doing a full run of livecheck with some additional temporary JSON information to see what strategies apply to the URL each check is currently using. I know some of the URLs only match one strategy but I need to review the ones where more than one strategy could apply and determine how to handle this situation.

Right now, the if/else block establishes a kind of precedence for strategies, so I need to see if any formulae rely on this specific strategy order and how we'll be able to deal with this as an ongoing concern. I have some ideas but I won't go into them until after I've reviewed the aforementioned data.

It may be that we won't be able to rework `#from_url` until after this PR and the subsequent one for the strategy feature are merged. If a check has more than one available strategy and only really works for one, we would want to add/update livecheckables with strategy information (e.g., `strategy :apache`) to ensure the check doesn't break when `#from_url`'s behavior changes.

Anyway, I'm making good progress and I'll get as much done in this PR as I can to enable taps to provide strategies but it's likely I'll have to create a follow-up PR to finish the work. Most of the hard work is already done, though.

maxim-belkin:
> Just to give you an update, I've done most of the work in the latest commit to enable taps to provide strategies. At the moment, all a tap would need to do is have a `livecheck_strategy` folder with strategies in them and they'll automatically load. Does this seem logical?

Thanks, Sam. Yes, it does seem logical and this is a step in the right direction! The actual directory name could use the word strategy in its plural form but that's completely unimportant.  

> [I'm still using `require_relative` statements for the built-in strategies here, until I can confirm whether the overly-clever way of requiring strategies (without being explicit) will work for documentation purposes.]

It's great that you're considering moving away from it. I think that homebrew-livecheck should be agnostic of strategy names internally. It should provide some strategies (most used, most useful), but it shouldn't be limited by our programming decisions (to use if/else block) for executing them.

I do have one more idea for you: consider allowing formulae to specify which strategy they want to use instead of auto-detecting it in `from_url`. If formulae provided the strategy they want to use, we wouldn't have to autodetect it. And by the way, you don't have to implement all of my ideas in this PR -- I just want you to be aware of them so that if you consider them worthy/reasonable you can make adjustments to your code.

> The tricky part was that `Homebrew.args.tap` wasn't available to `livecheck_strategy.rb`, since args aren't parsed until the main `livecheck` method is executed (i.e., it's not available when `livecheck_strategy.rb` is required). I worked around this by requiring the file after `livecheck_args.parse`, as I didn't find another working solution. It's not quite ideal but it works for now.

I'm glad you figured that out. This is a big change for taps. Previously, I had to use my hacked version of homebrew-livecheck that was rejected in #98. Please make sure to add a comment about the importance of order of things. It might be all we can do before homebrew-livecheck is merged into homebrew/brew so let's not spend too much time on that.

> That said, though the strategies are loaded and appear in `strategies`, they're not yet usable, as I still need to rework `#from_url`. I've already written an alternative implementation of this method that uses `strategies` (instead of the hardcoded if/else block) but I need to do some research before we can commit to using it.

That's great to hear! I'm very excited about this because ....  it will enable us to implement a scoring system for strategies (which can be based on various factors, such as amount it time/memory it takes, success rate, etc) and we could make adjustments to which strategy specific formula uses based on that scoring system. That's why I proposed enabling multiple strategies in #1010 and why I think it'll be useful to have even as a non-default behavior. When I tried it myself, I noticed that one strategy finished in a split second and the other one took 2 or 3 seconds but provided "more recent" version of an application. Right now strategies are implemented as "rules" that have their own non-overlapping regions of applicability. This is so because they use `url` to detect things. One could write an algorithm that uses something else, for example formula name.

Excellent work!

samford:
> The actual directory name could use the word strategy in its plural form but that's completely unimportant.

I agree that `livecheck_strategies` would be more evocative but I used `livecheck_strategy` to follow the existing naming convention in Homebrew/brew, where you have `unpack_strategy.rb` and an `unpack_strategy` folder containing the strategies. It's technically possible to allow usage of a `livecheck_strategies` name for taps only but I think it's better to maintain consistency. Taps appear to use the same folder names as the originals (like `Formula`) and I don't think we should go against that.

---

I tested whether the programmatic requiring of `.rb` files in homebrew-livecheck's `livecheck_strategy` folder worked with documentation generation and it does, so I restored it. The only reference to a specific strategy in `livecheck_strategy.rb` now is `:page_match` in `#from_url` and this is necessary to be able to conditionally include it in the usable strategies for a URL when a regex is provided.

The most recent commit contains full support for taps to provide their own strategies, as well as override the built-in ones. I've tested this out but I may have missed something unexpected, so feel free to kick the tires.

---

> I do have one more idea for you: consider allowing formulae to specify which strategy they want to use instead of auto-detecting it in `from_url`.

This is the main application of the `strategy` feature (née `using`), which I formalized in #301 on 2020-03-03 (after mentioning it in PR/issue comments for some time). I've lost count of how many times I've referenced the feature at this point, so I certainly agree that it will be useful once it's merged.

I've had the feature implemented and working for months and I've been extracting the changes into smaller, topical PRs (instead of one catch-all PR) as time permits. I wanted to create PRs sooner but being the primary mentor on our livecheck GSoC project takes up a lot of my free time usually. This is the final PR before the feature itself, which is partly why I've been wary of doing too much here and delaying it further.

[I've also been stashing changes for related livecheckables to create/modify once the feature's merged (and cataloguing other formulae to which it can be applied), so there will be also a lot of PRs for me to create once it's merged.]

---

As an aside, one thing that may be worth mentioning is that it's technically possible at this point to move livecheck blocks in livecheckables into formulae themselves. The established location is before `bottle` blocks and `brew audit` will let you know if it's misplaced.

We haven't done this here yet because we're going to revise the existing livecheckables to bring them all up to our current standards before the migration to formulae in homebrew-core (so they'll be representative examples for contributors). The clean-up work and migration is something that our GSoC student is working on (among other things).

Taps are free to make this change right now and get rid of the livecheckable files and `Livecheckables` directory. Unless there's something standing in the way, I would recommend that taps do this as soon as they can, as we will remove support for livecheckable files as part of the migration (i.e., by deleting livecheck's `Formulary` extension).

Thinking about it, I should create an issue about this, so folks can have something to see and reference. There's a chance no one will look at it but it's better than just suddenly pulling the rug out from under everyone else in the near future without warning.

---

I've looked through the checks for all the homebrew-core formulae at this point. In practical terms, the current behavior of the strategy order in `#from_url`'s if/else block can be reduced down to these rules:

* If the Git strategy is useable for a given URL, prefer it over other usable strategies.
* If the `page_match` strategy is usable for the URL (i.e., a regex is provided), only use it if there are no other usable strategies.

This is what I assumed would be the case but the data confirms that these two special cases are all we need to replicate the existing behavior.

I initially tested this by pooling the usable strategies for the URL and then selecting the strategy to use through a small if/else block. This worked but this behavior would have been hardcoded into livecheck, making it harder for taps to override it.

With this in mind, I added a `PRIORITY` constant to the `Git` and `PageMatch` strategies instead. I'm informally using a scale of 0 to 10, with 0 being lowest priority and 10 being highest. I assigned `Git` a priority of 8 and `PageMatch` a priority of 0, to establish the two rules mentioned above. [I used 8 for Git to allow some breathing room if another strategy needs a higher priority.] Any strategy without an established priority uses a default of 5 in `#from_url` (so they're in the middle, kind of like with the previous if/else block).

If a tap doesn't like the priority of the `Git` strategy, for example, they can create a `git.rb` file in their `livecheck_strategy` folder containing something like this:

```ruby
module LivecheckStrategy
  class Git
    PRIORITY = 10
  end
end
```

The `Git` strategy continues to function the same way but the priority is overridden.

All that said, I don't imagine we'll need a more complex priority system than this and I think this is likely flexible enough for our needs (as well as for taps). I'm reticent to introduce more complexity with a more complicated priority system, so I don't think that's something we should pursue unless there's a real need or clearly-demonstrated benefit.

---

One important thing to keep in mind when comparing sources/strategies for checks is that we prefer to align the source of the check (the URL) with the stable archive, whenever possible. I can't speak for taps but this is what we do here, at least. This is partly why I don't think executing multiple strategies is appropriate. In our context, we decide on one check that's most appropriate and use that.

The reason for this is because not all sources are equal. Sometimes the Git repo will tag a version but it's not a real release until it's on the upstream website. In this context, we check the upstream website and checking the Git repo would result in an erroneous result. We favor accuracy over considerations like performance (i.e., one source being faster than another) and the aforementioned preference is part of enforcing this.

maxim-belkin:
Thanks, Sam!
I agree with what you say above and I think this PR is 99.5% ready to be merged. Even though I do think that allowing execution of several strategies would've been useful, this is by no means is critical.

I've tested this PR already for formulae in Linuxbrew/xorg tap with https://github.com/Linuxbrew/homebrew-xorg/pull/659 and it works! I see that you use some fancy Ruby to call require method on strategy files. It might be useful to add debugging output  when files are loaded and which strategy is selected (`...[0]`). But other than that, this PR is a giant leap for Homebrew/livecheck! Great work! 🎉

livecheck/heuristic.rb
```diff
@@ -93,52 +63,41 @@ def latest_version(formula)
...
+    strategies = LivecheckStrategy.from_url(url, livecheck_regex.present?)
+    strategy = strategies[0]
 
     if Homebrew.args.debug?
-      puts "\nURL:             #{original_url}"
-      puts "URL (processed): #{url}" if url != original_url
-      puts "Strategy:        #{strategy.nil? ? "None" : strategy.to_s.delete_suffix("_strategy")}"
-      puts "Regex:           #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
+      puts "URL (processed):  #{url}" if url != original_url
+      puts "Strategies:       #{strategies.map { |s| s.name.demodulize }}" unless strategies.empty? || !Homebrew.args.verbose?
```
maxim-belkin, line 86: Currently it produces output like this:
```
Strategies:       ["PageMatch"]
```
I suggest making it a bit more human-readable:
```suggestion
      puts "Strategies:       #{strategies.map { |s| s.name.demodulize }.join(", ")}" unless strategies.empty? || !Homebrew.args.verbose?
```

samford: I can get behind this. I need to make a change to this line to fix a style issue, so I'll incorporate it in a reworking of the latest commit.

samford:
> Even though I do think that allowing execution of several strategies would've been useful, this is by no means is critical.

I can be convinced otherwise if I'm presented with compelling evidence of the benefit of allowing it in certain circumstances but it's a hard sell for me thus far, especially if it's applied to all checks. To me, the examples in your PR showed additional effort without any real benefit. If you plan to pursue this further, it may be better to open an issue to discuss how it can be approached and to determine how we could apply it before moving forward with a PR.

That said, I modified `#from_url` to return all usable strategies and moved the `usable_strategies[0]` behavior into `latest_version()`, where we call `#from_url`. This makes it possible to print the available strategies for a URL in the debug output (I put it behind the `verbose` flag for now, as I don't think it's useful enough to always be included in debug output) but it technically also makes it easier to enable multiple strategies in some regard. However, this is intended to improve the debug output and know that it's not tacit approval of multiple strategies.

Related, I made some improvements to the verbose debug output to make it more usable (most importantly, removing the lengthy `$LOAD_PATH` output). This also includes an `odebug` call to alert the user when a GitHub Gist URL is skipped.

I also added the available strategies for the URL used in the check to the verbose JSON output. I'm omitting it if it's empty and I'll update it if we decide to avoid conditional inclusion of JSON information in the forthcoming issue to discuss this.

> It might be useful to add debugging output when files are loaded and which strategy is selected (`...[0]`).

This _may_ be useful to include in the verbose debug output but it would seemingly require creating a method that prints a loading message (like `Formula.load_formula_from_path`) before a given strategy file is required. I don't think it's hard or would require much code but I don't want to delay this PR in working on it right this moment. I'll try to work on this in a follow-up PR.

---

At this point, I'm going to go over the changes and do any final polishing that remains but I don't imagine there's much else that would be appropriate to do here (instead of in subsequent PRs).

I really appreciate all the suggestions, feedback, and discussion and the PR's definitely better from all this review. Thanks, Maxim!

livecheck/heuristic.rb
samford: I added some `odebug` output for this situation but I haven't yet looked into whether there's a better way of dealing with this situation. This is fine for now, so I'm going revisit this down the road, after I've finished working on some things that have broader applications.

maxim-belkin:
Thanks, Sam. Just a very quick note before you merge this: it looks like there is a regression somewhere: when I run `brew livecheck -d --tap linuxbrew/xorg` some formulae return "Unable to get versions" but when I check them manually they work just fine. I need to check why that happens.

samford:
Sure, I'm not ready to merge this just yet and I'm planning to leave this open for a while longer anyway (in case anyone else wants to take a last pass at reviewing it). I'm still reviewing a full before/after livecheck run over the homebrew-core formulae to spot regressions as well, so let me know what you find in Linuxbrew/homebrew-xorg.

maxim-belkin:
I figured out what the problem is. Tap strategies are loaded only when you use `--tap` option. When you specify formula by name, the default strategies are used. I suggest determining the taps that specified formulae originate from (e.g., using the syntax below) and then using it instead of `Homebrew.args.taps` in livecheck/livecheck_strategy.rb:

```rb
extra_taps = formulae_to_check.map { |f| f.tap.name }.uniq - ["homebrew/core"]
```

maxim-belkin:
Also, I wonder if we should use `full_name` for formulae names in cmd/livecheck.rb to make sure we process the right one (e.g. when a tap has a formula that replicates the one from homebrew/core but with some modifications (e.g. options)).

samford:
> Tap strategies are loaded only when you use `--tap` option.

Ahh, I thought I had tested that situation but I clearly must have only tested `--tap`. It makes total sense that it would fail like this.

I moved the tap-strategy-requiring logic into `Homebrew#livecheck` (since that's where `formulae_to_check` exists) and this seems to work as expected. It also allows us to move the `require_relative` statement for `livecheck_strategy.rb` out of `Homebrew#livecheck`, so that's an added bonus.

Let me try to finesse this a little and I'll push a commit when I'm done. Thanks for bringing this to my attention and for the snippet.

samford:
> I wonder if we should use `full_name` for formulae names in cmd/livecheck.rb

Seems like it. With `Formula.names`, only the homebrew-core formula is loaded when there's a name overlap but `Formula.full_names` loads both. I assume you're referencing the code in the `Homebrew.args.all?` conditional block but was there anywhere else where the change needed to be made as well?

maxim-belkin:
> I moved the tap-strategy-requiring logic into `Homebrew#livecheck` (since that's where `formulae_to_check` exists) and this seems to work as expected. It also allows us to move the `require_relative` statement for `livecheck_strategy.rb` out of `Homebrew#livecheck`, so that's an added bonus.
> 
> Let me try to finesse this a little and I'll push a commit when I'm done. Thanks for bringing this to my attention and for the snippet.

No problem! Thanks for fixing it 👍 


>  I assume you're referencing the code in the `Homebrew.args.all?` conditional block

Yes, that's the one! 👍

maxim-belkin:
```
brew livecheck -vd wtf
```
produces:
```
Error: wtf: undefined method `except' for nil:NilClass
```
that comes from `usable_strategies = strategies.except(:page_match).values.select do |strategy|` in `livecheck/livecheck_strategy.rb`

samford:
Assuming this is the homebrew-core `wtf`, it works fine for me on this branch as it exists here. I can't imagine how `strategies` could be `nil`...

maxim-belkin:
Let me check... I did add some additional output...

maxim-belkin:
my mistake -- added `odebug "Computed #{@strategies}"` to the end of `def self.strategies` which caused that `nil`.

livecheck/livecheck_strategy/apache.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module LivecheckStrategy
+  class Apache
+    NICE_NAME = "Apache"
+    NAME = NICE_NAME.downcase
```
maxim-belkin, line 6: I believe NICE_NAME and NAME aren't necessary now, are they?

maxim-belkin:
Currently, `page_matches` (in livecheck/utils.rb) can not get data from FTP websites. There is a handful of formulae that are affected. To allow FTP sources:

```gitdiff
diff --git a/livecheck/utils.rb b/livecheck/utils.rb
index d63d8d4..a037454 100644
--- a/livecheck/utils.rb
+++ b/livecheck/utils.rb
@@ -1,4 +1,5 @@
 require "open-uri"
+require "net/ftp"
 
 def checkable_urls(formula)
   urls = []
@@ -27,7 +28,17 @@ def git_tags(repo_url, filter = nil)
 end
 
 def page_matches(url, regex)
-  page = URI.open(url).read
+  entry = URI.parse(url)
+  scheme = entry.scheme
+  if scheme.start_with?("http")
+    page = entry.read
+  elsif scheme.start_with?("ftp")
+    ftp = Net::FTP.new(entry.host)
+    ftp.login
+    ftp.chdir(entry.path)
+    page = ftp.list('*').join("\n")
+  end
+
   matches = page.scan(regex)
   matches.map(&:first).uniq
 end
```

I've also come up with a last-resort strategy (I called it Default but it can be called fallback or last_resort) that can be attempted with various packages when everything else fails. :)
It helps check versions for snobol4, openconnect, cpl, maybe some other packages

```rb
# frozen_string_literal: true

module LivecheckStrategy
  class Default
    PRIORITY = 0
    def self.match?(url)
      true
    end

    def self.find_versions(url, regex)
      components = url.split("/")
      page_url = components[0..-2].join("/")
      package_name = components[-1].split("-")[0..-2].join("-")
      puts "Package_name", package_name
      regex ||= /#{package_name}-([\d\w]+(?:[\.\d\w]+)+)(?:\.tar\.[\d\w]*)|(?:\.tgz)/
      PageMatch.find_versions(page_url, regex)
    end
  end
end
```

livecheck/livecheck_strategy/apache.rb
samford: This is one of the things I was going to reconsider when I take a final clean-up pass at this.

Since we use `strategy.name.demodulize` a few times in `latest_version()`, it seems like it would be better to update `NAME` to `name.demodulize` (executing it only once for each strategy) and go back to using `strategy::NAME` instead. With this setup, any taps would simply need to make sure to use `NAME = name.demodulize` in their strategies to follow suit.

Ideally, we would be able to automatically provide this constant to all strategies without requiring manual duplication of this code but Ruby doesn't allow us to dynamically assign constants. We also never instantiate these classes, so we can't use `initialize` for this either.

It's not ideal but I don't think it's a lot to ask from taps and it saves us a little effort in `latest_version()`.

---

There _may_ (or may not) be a use for `NICE_NAME` in the future but there isn't one currently. It's only useful for cases like `Sourceforge`, `Pypi`, `Gnu`, etc. anyway, where the `NICE_NAME` (`SourceForge`, `PyPI`, `GNU`) is prettier than the class name.

The reason why we don't use it in the current debut output is because we've standardized on using the class name. Using the `NICE_NAME` in debug output could confuse users (i.e., "The output says `SourceForge` but the class name is actually `Sourceforge`"), so it may end up that `NICE_NAME` can never be appropriately used. For the time being, I'm keeping it in the strategies where it differs from the class name. It will be there if we need to use it and we can always remove it in the future if we change our mind.

[For what it's worth, the reason why it's necessary to use the `Sourceforge` and `Pypi` names for the classes is so that we get the proper `:sourceforge` and `:pypi` symbols when `@strategies` is generated. If the class uses the prettier names, we get `:source_forge` and `:py_pi`, which isn't so nice.]

samford:
> allow FTP sources

I previously created an `ftp_matches()` function [in response to the open `pstree` PR](https://github.com/Homebrew/homebrew-livecheck/pull/829#issuecomment-643689884) but I haven't decided whether this is an appropriate check to make yet. It's technically possible but checking an FTP server is a little more involved than fetching text sources (HTML/RSS/JSON/etc.), so I want to be sure it won't be a burden for FTP servers.

I probably need to take another pass or two at this but my WIP implementation looks like this:

```ruby
def ftp_matches(url, regex = nil)
  uri = URI.parse(url)
  host = uri.host
  path = uri.path

  return nil if host.nil? || host.empty?

  files = Net::FTP.open(host) do |ftp|
    ftp.login
    ftp.nlst(path)
  end

  unless regex.nil?
    files = files.map do |path|
      file = Pathname(path).basename.to_s
      matches = regex.match(file)

      next if matches.nil? || matches.length < 1
      matches[1]
    end.compact
  end

  files.map { |f| Version.new(f) }
end
```

I tried to reduce it down to the fewest number of FTP commands for the sake of efficiency.

If you think having an FTP strategy would be useful (e.g., it's currently the only way to create a check for `pstree`), then I'll work this into a new `Ftp` strategy in a follow-up PR. I may omit it from being included in `usable_strategies` and instead require that a livecheckable with an explicit strategy (`strategy :ftp`) be created. Until I'm convinced this won't be a burden for upstream FTP servers, I don't want to use it widely.

---

What you're doing in the `Default` strategy here is something that I've been working on as well. We have a lot of checks where we use the URL without the file name (e.g., `https://www.example.com/files/example-1.2.3.tar.gz` -> `https://www.example.com/files/`), so I've been exploring ways of addressing this. Ideally, it would be nice to simply be able to use `url :stable` instead of the hardcoded URL in these cases.

However, I'm not yet sold on the idea of having a default strategy. I'm of the opinion that it's preferable for livecheck to fail when no strategies apply, rather than potentially wasting effort on a default strategy that _also_ may not work. With the current setup, we simply identify failing checks, add a livecheckable to create a working check, and livecheck doesn't waste effort in the interim time. [An aside: I have our GSoC student working on creating livecheckables for the formulae with errors (~650, down from 900), so we'll eventually have working checks for nearly all formulae.]

Instead, I think it may be more appropriate to achieve the same goal you're aiming at with a different approach. We want to get the benefits (using `url :stable`) but without the detriments (introducing wasted effort).

My initial thinking on how to achieve this is to simply remove the file name from the URL when it's an archive file (i.e., we don't want to remove .html, .json, etc.). This would be appropriate to do in `preprocess_url`, like we do with GitHub and GitLab URLs.

I did some early work and when I last lefit it I was working on file type detection. I had written a script to go through all the stable URLs in formulae and list the file extensions, so I could see if `UnpackStrategy` would cover everything. I think it's probably reasonable to start with using `UnpackStrategy`, even if it doesn't cover everything, and this would at least enable us to use it for the file types it does cover.

Doing things in this way means that `url :stable` would work and it wouldn't introduced any wasted effort because `PageMatch` is only executed when a `regex` is provided (i.e., checks without a livecheckable will continue to not match any strategies and not waste effort).

This is outside of the scope of this PR as well but I'll create a follow-up PR once this one is merged and we can continue the discussion there.

---

That said, I would appreciate if we could try to limit feedback/discussion to items that are appropriate to work on here. At this point we should be focusing on clean-up rather than feature additions. No doubt, there's plenty of room for improvement in homebrew-livecheck but we can't (and shouldn't) do it all in one PR.

maxim-belkin:
Hi, Sam!

> I previously created an `ftp_matches()` function

That's great! I think that enabling FTP sources is important (I'm not aware of any restrictions that would prevent people from creating formulae for packages stored on FTP). But I'm not sure where anyone could find this implementation. It looks like you do a lot of work that stays on your computer... or at least leaves there for a while before you create a PR.

As for the actual implementation -- I'm sure there is a reason why you implemented it this way but I think you could instead merge the list of files into `page` variable and then hand it off to the common http/ftp block (`matches = page.scan(regex)`). But this is a technical detail.

> Until I'm convinced this won't be a burden for upstream FTP servers, I don't want to use it widely.

I think you're over-complicating things. I can't imagine a handful of requests to cause problems for a server.

> What you're doing in the `Default` strategy here is something that I've been working on as well. We have a lot of checks where we use the URL without the file name (e.g., `https://www.example.com/files/example-1.2.3.tar.gz` -> `https://www.example.com/files/`), so I've been exploring ways of addressing this. Ideally, it would be nice to simply be able to use `url :stable` instead of the hardcoded URL in these cases.
> 
> However, I'm not yet sold on the idea of having a default strategy. I'm of the opinion that it's preferable for livecheck to fail when no strategies apply, rather than potentially wasting effort on a default strategy that _also_ may not work.

This is where I have a different opinion. I think there should be some "default" strategy applicable to a wide variety of cases/formulae and then special cases handled by the additional strategies (such as those that we have here -- Git, SourceForge, etc.) when the default one "fails" or finds versions that are "too recent". That default strategy would let us use  `livecheck_strategy "default"` (or something like that) in formulae instead of a full-fledged `livecheck do` block. 
I didn't quite understand what you said about `url :stable`.

samford:
In light of the shortened timeline for the homebrew-core and Homebrew/brew migrations (as mentioned in our internal conversation), I'm taking a final clean-up pass at this. Afterward, I'll test it to make sure it's functionally identical to the main branch (outside of the additional JSON values) before merging.

We can revisit the idea of an FTP strategy sometime after the Homebrew/brew migration is finished but I won't have time until sometime after that. However, it may be worth mentioning that I'm not married to my implementation and I think there's value in discussing how we should approach it (i.e., what the benefits/detriments of each approach are, what's appropriate, etc.). The solution may end up being somewhere between our two approaches but we'll figure that out through discussion in the future.

Similarly, I'm happy to discuss the idea of whether a default strategy is appropriate or beneficial sometime after the migration. As with any notable change, it would be worth creating an issue at that point (not now) to make the case for the change. If you're looking to apply it in a broad sense, it would be necessary to demonstrate that it's clearly beneficial across the board and doesn't come with undesirable side effects. I'm sure I'll do my own research and testing as well and we can collectively determine the best way forward based on the evidence in front of us.

Thanks again for all the feedback, suggestions, and discussions here!

samford:
After some final clean-up, I confirmed that this all works as expected across all the core formulae (homebrew/core and linuxbrew/core). Notably, the strategies still function the same way after the refactoring and the selected strategy for each URL remains the same after the addition of the priority setup.

This is everything that I wanted to cover in this PR, so I'm going to go ahead and merge it now.

Issue #1036, opened 2020-06-28
nomad 0.12.x is in beta


chenrui333:
stable release listed in here, https://www.nomadproject.io/downloads

## Current behavior

```
$ brew livecheck -v nomad
nomad : 0.11.3 ==> 0.12.0
```

## Expected behavior

```
$ brew livecheck -v nomad
nomad : 0.11.3 ==> 0.11.3
```

PR #1039, opened 2020-06-29
nomad: update livecheckable

miccal:
I have updated the `livecheckable` for `nomad` to use `https://releases.hashicorp.com/nomad/` for the `url` and added `[<\/a>]` to the regex based on studying the page source of the `url` -- this is so that things like `-beta1`, `-rc1` and/or `-connect1` are not included in `livecheck`.

Output before:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck nomad
nomad : 0.11.3 ==> 0.12.0
```

Output after:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck nomad
nomad : 0.11.3 ==> 0.11.3
```

Closes https://github.com/Homebrew/homebrew-livecheck/issues/1036

Thanks.

**EDIT:** This is what I had originally:
```
class Nomad
  livecheck do
    url "https://releases.hashicorp.com/nomad/"
    regex(/href=.*?nomad_v?(\d+(?:\.\d+)+)[<\/a>]/i)
  end
end
```

### DIFF
--- a/Livecheckables/nomad.rb
+++ b/Livecheckables/nomad.rb
@@ -1,6 +1,5 @@
 class Nomad
   livecheck do
-    url "https://www.nomadproject.io/downloads.html"
-    regex(%r{href="https://releases.hashicorp.com/nomad/([0-9.]+)})
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
   end
 end
--- a/Livecheckables/nomad.rb
+++ b/Livecheckables/nomad.rb
@@ -1,5 +1,6 @@
 class Nomad
   livecheck do
+    url :head
     regex(/^v?(\d+(?:\.\d+)+)$/i)
   end
 end


miccal:
Alternatively, since `nomad` uses GitHub for their releases we could inetead just use:
```
class Nomad
  livecheck do
    regex(/^v?(\d+(?:\.\d+)+)$/i)
  end
end
```

miccal:
```
C:  4: 11: Use %r around regular expression.
```
Not sure what I have done wrong here?

Livecheckables/nomad.rb
```diff
@@ -1,6 +1,5 @@
 class Nomad
   livecheck do
-    url "https://www.nomadproject.io/downloads.html"
-    regex(%r{href="https://releases.hashicorp.com/nomad/([0-9.]+)})
+    regex(/^v?(\d+(?:\.\d+)+)$/i)
```
nandahkrishna, line 3: I'm currently working on ensuring all livecheckables have a `url` in them, so it would be great if you can add the URL here too @miccal.

```suggestion
    url :head
    regex(/^v?(\d+(?:\.\d+)+)$/i)
```

miccal: Thanks for your review @nandahkrishna.

samford:
Yep, you got it! Since the stable archive comes from GitHub, we should be checking the Git repo tags like you're doing here.

PR #1042, opened 2020-06-29
cairo: update livecheckable

miccal:
I have slightly updated the `livecheckable` for `cairo` to conform to the recent style, combined with the fact that only even minor versions are stable, as explained [here](https://www.cairographics.org/manual/cairo-Version-Information.html) -- I had to add `[^r]` to ensure that the releases for `rcairo` were not included. This matches https://github.com/Homebrew/homebrew-livecheck/pull/1041, and it outputs the same as before:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck cairo
cairo : 1.16.0 ==> 1.16.0
```

Thanks.

### DIFF
--- a/Livecheckables/cairo.rb
+++ b/Livecheckables/cairo.rb
@@ -1,6 +1,6 @@
 class Cairo
   livecheck do
-    url "https://www.cairographics.org/news/"
-    regex(/cairo-([0-9]+\.[0-9]*[02468]\.[0-9]+)/)
+    url "https://cairographics.org/releases/?C=M&O=D"
+    regex(/href=.*?[^r]cairo-v?(\d+\.\d*[02468](?:\.\d+)*)\.t/i)
   end
 end
--- a/Livecheckables/cairo.rb
+++ b/Livecheckables/cairo.rb
@@ -1,6 +1,6 @@
 class Cairo
   livecheck do
     url "https://cairographics.org/releases/?C=M&O=D"
-    regex(/href=.*?[^r]cairo-v?(\d+\.\d*[02468](?:\.\d+)*)\.t/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/cairo.rb
```diff
@@ -1,6 +1,6 @@
 class Cairo
   livecheck do
-    url "https://www.cairographics.org/news/"
-    regex(/cairo-([0-9]+\.[0-9]*[02468]\.[0-9]+)/)
+    url "https://cairographics.org/releases/?C=M&O=D"
+    regex(/href=.*?[^r]cairo-v?(\d+\.\d*[02468](?:\.\d+)*)\.t/i)
```
samford, line 4:
```suggestion
    regex(%r{href=(?:["']?|.*?/)cairo-v?(\d+\.\d*[02468](?:\.\d+)*)\.t}i)
```

miccal: Ah, a much better and future-proof way, thanks @samford. I missed the `pycairo` matches, apologies for that oversight.

miccal: @samford a question if I may -- is the _only_ reason you use `%r{..}` as opposed to `\..\` so that you do not have to escape the `/` in your `(?:["']?|.*?/)`-bit, or are there other reasons?

samford: Yep, it's so we don't have to escape forward slashes in the regex. `brew style` enforces use of the `/.../` regex literal syntax when the regex doesn't include `/` and the `%r{...}` regex literal syntax when it does include `/`.

miccal: Thank you as always for your explanation (and patience).

Issue #1040, opened 2020-06-29
cairomm stable release is still 1.12.2


chenrui333:
https://github.com/Homebrew/homebrew-core/pull/37287#issuecomment-466827793

> 1.12.2 is still marked as the latest stable version of cairomm in https://github.com/freedesktop/cairomm/blob/master/NEWS

## Current behavior

```
$ brew livecheck -v cairomm
cairomm : 1.12.2 ==> 1.15.5
```

## Expected behavior

```
$ brew livecheck -v cairomm
cairomm : 1.12.2 ==> 1.12.2
```

PR #1043, opened 2020-06-29
cairomm: update livecheckable

miccal:
I have updated the `livecheckable` for `cairomm` to conform to the recent style, combined with the assumption that only even minor versions are stable, as explained [here for `cairo`](https://www.cairographics.org/manual/cairo-Version-Information.html). This matches https://github.com/Homebrew/homebrew-livecheck/pull/1041  and https://github.com/Homebrew/homebrew-livecheck/pull/1042.

Output before:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck cairomm
cairomm : 1.12.2 ==> 1.15.5
```

Output after:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck cairomm
cairomm : 1.12.2 ==> 1.12.2
```

Closes https://github.com/Homebrew/homebrew-livecheck/issues/1040

Thanks.

### DIFF
--- a/Livecheckables/cairomm.rb
+++ b/Livecheckables/cairomm.rb
@@ -1,6 +1,6 @@
 class Cairomm
   livecheck do
-    url "https://cairographics.org/releases/"
-    regex(/href=.*?cairomm-v?(\d+(?:\.\d+)+)\.t/i)
+    url "https://cairographics.org/releases/?C=M&O=D"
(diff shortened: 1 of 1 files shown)


miccal:
@samford, yes I noticed that too -- hence why I updated all three of the `livecheckable`'s in case they do this in the future.

PR #1051, opened 2020-07-01
mockserver: add livecheckable

nandahkrishna:
Adding Livecheckable for `mockserver`.

### DIFF
--- a/Livecheckables/mockserver.rb
+++ b/Livecheckables/mockserver.rb
@@ -1,6 +1,6 @@
 class Mockserver
   livecheck do
-    url "https://repo1.maven.org/maven2/org/mock-server/mockserver-netty/"
+    url "https://oss.sonatype.org/content/repositories/releases/org/mock-server/mockserver-netty/"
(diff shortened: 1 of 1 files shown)


nandahkrishna:
There seems to be an issue with this one, there's a version `5.10` released at 07:06 on 24-03-2020 and a `5.10.0` released at 07:32 on 24-03-2020. Here's the livecheck output:
```
mockserver : 5.9.0 => 5.10
```
The MD5 of the `5.10` and `5.10.0` versions aren't the same. How should we handle this one?

miccal:
Perhaps `https://github.com/mock-server/mockserver.git` might be better?

nandahkrishna:
I think in this case, as we're using the archive from 
```
https://oss.sonatype.org/content/repositories/releases/org/mock-server/mockserver-netty/
5.9.0/mockserver-netty-5.9.0-brew-tar.tar
```
in the Formula, this url might be better. I thought we'd rather not check a url that's not mentioned in the Formula. Of course, thanks for the input @miccal and we might consider this option too based on what @samford thinks.

Issue #1049, opened 2020-07-01
mit-scheme stable release issue


chenrui333:
mit-scheme stable release url, https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/

## Current behavior

```
$ brew livecheck -v mit-scheme
mit-scheme : 10.1.11 ==> 11.0.90
```

## Expected behavior

```
$ brew livecheck -v mit-scheme
mit-scheme : 10.1.11 ==> 10.1.11
```

PR #1052, opened 2020-07-01
mit-scheme: update livecheckable

miccal:
The `:homepage` lists versions for both the stable and testing releases, so I have changed the `url` as suggested in https://github.com/Homebrew/homebrew-livecheck/issues/1049 and updated the regex accordingly.

Output before:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck mit-scheme
mit-scheme : 10.1.11 ==> 11.0.90
```

Output after:
```
-bash-5.0.17- /Users/miccal (29) [> brew livecheck mit-scheme
mit-scheme : 10.1.11 ==> 10.1.11
```

Closes https://github.com/Homebrew/homebrew-livecheck/issues/1049

Thanks.

### DIFF
--- a/Livecheckables/mit-scheme.rb
+++ b/Livecheckables/mit-scheme.rb
@@ -1,6 +1,6 @@
 class MitScheme
   livecheck do
     url "https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/?C=M&O=D"
-    regex(/href=.*?(\d+(?:\.\d+)+)/)
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?[ '">]}i)
   end
 end
--- a/Livecheckables/mit-scheme.rb
+++ b/Livecheckables/mit-scheme.rb
@@ -1,6 +1,6 @@
 class MitScheme
   livecheck do
-    url :homepage
-    regex(/href=.*?mit-scheme-v?(\d+(?:\.\d+)+)\.t/i)
+    url "https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/?C=M&O=D"
+    regex(/href=.*?(\d+(?:\.\d+)+)/)
   end
 end


miccal:
I think the `i` at the end of the regex is not really necessary here?

nandahkrishna:
> I think the i at the end of the regex is not really necessary here?

I believe it's better to keep things case-insensitive, just in case the check fails some time in the future because someone decided to make it `MIT-Scheme` rather than `mit-scheme` or something of that sort. I believe that was @samford's intention when they suggested making the regexes case insensitive, hope I'm making sense here.

chenrui333:
@miccal I am going to close your PR for now, since homebrew core team is going to focus the energy to make the livecheck flow better and we can make this effort moving forward at large. 

Honestly the issues that I created mostly serve for my own sake, and I can quickly implement those solutions in my branch. (we will revisit after some time). Have chatted with Sam privately on this matter.

Thanks a lot for your contributions though.

cc @samford

samford:
There was some confusion here. I'm going to take the time to merge this when I get a chance. We had a discussion about minimizing _new_ work (issues/PRs) over the coming weeks while we focus on work related to migrating homebrew-livecheck to homebrew-core and Homebrew/brew. I'm going to take the time to merge the open PRs when I have a free moment.

I'm working to finish internal work that blocks the Google Summer of Code (GSoC) Homebrew/brew migration by July 13 and I won't have as much time to deal with unrelated PRs over the next week and a half. This is a temporary situation, so I had talked to @chenrui333 about holding off on creating new issues/PRs that end up directing work towards livecheckable contributions. We'll address these issues in the future but now isn't the best time for it.

I appreciate everyone's understanding during this period.

Livecheckables/mit-scheme.rb
```diff
@@ -1,6 +1,6 @@
 class MitScheme
   livecheck do
-    url :homepage
-    regex(/href=.*?mit-scheme-v?(\d+(?:\.\d+)+)\.t/i)
+    url "https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/?C=M&O=D"
+    regex(/href=.*?(\d+(?:\.\d+)+)/)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?[ '">]}i)
```

miccal: Thanks @samford.

PR #1063, opened 2020-07-05
Change \.z to \.zip in regex

nandahkrishna:
In this PR, Livecheckable regexes containing `\.z` are modified to use `\.zip`. The Livecheckables affected are:
```
1 antlr4-cpp-runtime.rb
2 exomizer.rb
3 groovy.rb
4 groovysdk.rb
5 gsoap.rb
6 ptex.rb
7 sonarqube.rb
8 spring-roo.rb
9 wirouter_keyrec.rb
```


samford:
I tested the modified formulae here both before and after the changes and the output is identical, as expected.

PR #1078, opened 2020-07-05
Allow to livecheck formulae without tap

Moisan:
Formula.tap can be nil in some cases. Currently this make livecheck fail without any useful error message. I think the simplest fix is to skip those when looking at non-core taps.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -83,9 +83,7 @@ module Homebrew
     # Identify any non-homebrew/core taps in use for current formulae
     non_core_taps = {}
     formulae_to_check.each do |f|
-      unless f.tap.nil?
-        non_core_taps[f.tap.name] = true unless f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
-      end
+      non_core_taps[f.tap.name] = true unless f.tap.nil? || f.tap.name == "homebrew/core"
     end
     non_core_taps = non_core_taps.keys.sort
 
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -83,7 +83,9 @@ module Homebrew
     # Identify any non-homebrew/core taps in use for current formulae
     non_core_taps = {}
     formulae_to_check.each do |f|
-      non_core_taps[f.tap.name] = true unless f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
+      unless f.tap.nil?
+        non_core_taps[f.tap.name] = true unless f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
+      end
     end
     non_core_taps = non_core_taps.keys.sort
 


cmd/livecheck.rb
```diff
@@ -83,7 +83,9 @@ def livecheck
     # Identify any non-homebrew/core taps in use for current formulae
     non_core_taps = {}
     formulae_to_check.each do |f|
-      non_core_taps[f.tap.name] = true unless f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
+      unless f.tap.nil?
+        non_core_taps[f.tap.name] = true unless f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
+      end
```
samford, line 88:
```suggestion
      unless f.tap.nil? || f.tap.name == "homebrew/core" || non_core_taps.key?(f.tap.name)
        non_core_taps[f.tap.name] = true
      end
```

Since this only has one line, it seems like we can potentially tidy it up a bit by moving the other conditions into the outer `unless`.

Bo98: To me, it seems like that whole `unless` line can be simplified to `unless f.tap&.name == "homebrew/core"`

samford: Ah, thanks for the reminder. That was the other approach I was going to suggest (the safe navigation operator) but I completely forgot by the time I got to the computer, ahah.

I confirmed that this works fine for me locally (i.e., the safe navigation operator takes the place of `f.tap.nil?` and we're fine to remove the additional `|| non_core_taps.key?(f.tap.name)` condition), so I updated the suggestion above.

Moisan:
One example I have locally is `python@2`: I installed a few formulae that depend on it in the past and I didn't run `brew cleanup` in a while, so it is still present. Since it was on homebrew-core but now isn't, there is no tap associated to it.

cmd/livecheck.rb
Moisan: This change still creates an error since it's accessing the `f.tap.name` at the beginning of the line. Hence I get 
```
Error: undefined method `name' for nil:NilClass
```

samford:
Ah, thanks for that. I'm able to replicate it now by installing a test formula, temporarily deleting the formula, and then running livecheck on that formula.

cmd/livecheck.rb
samford: Ahh, yeah, `f.tap&.name` will be `nil` in this case and `nil` technically doesn't equal `"homebrew/core"`, so it passes through.

Looks like `f.tap.nil?` is the only way to go after all. The current suggestion removes the unnecessary `|| non_core_taps.key?(f.tap.name)`, which lets us maintain the inline conditional. I've tested this locally and it works as expected for a formula without a tap.

SMillerDev:
An easier version might be installing something from a seperate formula file

samford:
> An easier version might be installing something from a seperate formula file

I wasn't aware this was an option but I put together a simple test formula outside of a tap for this and now I can install it when I need to test this and uninstall it afterward. I appreciate the heads up.

---

If we're good with the suggestion above, I'm happy to commit it and merge this PR.

PR #1079, opened 2020-07-07
README: update example livecheckable formula

SeekingMeaning:
The example in README.md is outdated and doesn't match the current format/style

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,10 @@ Here's an example from [python.rb](Livecheckables/python.rb):
 
 ```ruby
 class Python
-  livecheck :url   => "https://www.python.org/downloads/",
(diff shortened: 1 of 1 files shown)


SeekingMeaning:
(Don't forget to update the link right above the code block)

SeekingMeaning:
Ah okay, no worries. Thank you for your work! (I genuinely mean it)

samford:
Hopefully I'll see you again on the other side of these migrations. By that point, we should also have some good documentation that thoroughly explains everything (e.g., how livecheck works, the `livecheck` block format, `url` and `regex` guidelines, strategy information, etc.), so keep an eye out for that. Until then!

PR #1085, opened 2020-07-10
juju: update url and regex

chenrui333:
```
$ brew livecheck -v juju
juju : 2.8.1 ==> 2.7.7
```

relates to https://github.com/Homebrew/homebrew-core/pull/57798

### DIFF
--- a/Livecheckables/juju.rb
+++ b/Livecheckables/juju.rb
@@ -1,6 +1,6 @@
 class Juju
   livecheck do
-    url :stable
-    regex(/^juju-v?(\d+(?:\.\d+)+)$/i)
+    url "https://github.com/juju/juju/releases/latest"
+    regex(%r{href=.+?/tag/juju-v?(\d+(?:\.\d+)+)}i)
   end
 end


chenrui333:
Looks like there are some discussions around whichever is the latest.

chenrui333:
`/latest` wont fit. will discuss a better solution on this.

samford:
@chenrui333 From livecheck's perspective, if they publish pre-release versions that use the same tag format as stable releases (i.e., `juju-2.8.0` is the latest stable but `juju-2.8.1` is currently prerelease on GitHub), then there's no way for us to identify the newest stable release if "latest" on GitHub isn't actually the latest stable release.

The "latest" marker is determined by the timestamp of the tag, so upstream would need to modify the date of the `juju-2.8.0` tag to be newer than the `juju-2.7.7` tag for "latest" to be accurate. I think it's appropriate for livecheck to check the "latest" release and upstream just needs to fix their tags and be careful about how they deal with release dates in the future.

Due to the nature of the upstream situation, there's nothing we can do in livecheck to address this.

chenrui333:
Yeah, I already asked if they can modify their release tag strategy, I can for sure this is problematic for the future.

PR #1088, opened 2020-07-11
Integrate heuristic and utils into livecheck.rb

samford:
The main goal of this PR was to delete `livecheck/heuristic.rb` by moving it somewhere else. This is an important part of the refactoring work to get homebrew-livecheck into a state where it can be integrated into Homebrew/brew without too much trouble.

In this case, I integrated `heuristic.rb` and `utils.rb` into `cmd/livecheck.rb`. This will allow us to delete the `livecheck` directory during the Homebrew/brew migration, since we'll be removing the `Formulary` extension at that time.

It may be that this will be reworked again during the Homebrew/brew migration but this should be a better starting point, at least.

---

This PR includes the following changes:

* Move `heuristic.rb` constants and functions into `cmd/livecheck.rb`
* Move `utils.rb` functions into `cmd/livecheck.rb`
* Delete `heuristic.rb` and `utils.rb`

* Simplify `#print_latest_version` by moving unrelated logic into `#livecheck`
  * Use the "JSON" hash to store the version information (even when we're not outputting JSON)
  * Rework `#print_latest_version` to output information using the hash info
  * Move formula-skipping code into a separate function (`#skip_conditions`)

* Create a function that returns "status" Hash objects (`#status_hash`) to reduce repetition

* Create additional constants for duplicated strings
* Move top-level constants in `cmd/livecheck.rb` into the `Homebrew` module
* Make internal constants and functions private

---

I need to have all the refactoring work done by the end of tomorrow, so I don't want to accomplish much more than this here. However, I would appreciate some review on this, to get some feedback on the approaches here and to make sure I didn't make any simple mistakes.


cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
-      version_info.nil? ? nil : version_info["latest"]
-    else
-      formula.head.downloader.fetch_last_commit
-    end
+    current_s =
+      if info["version"]["newer_than_upstream"]
+        "#{Tty.red}#{info["version"]["current"]}#{Tty.reset}"
+      else
+        info["version"]["current"]
+      end
```
maxim-belkin, line 282: 1. I understand that you're preserving the old formatting but I wonder if it's worth it given that it adds a conditional block.
2. If case you think these conditionals should be kept -- can you combine them? Both `info["version"]["newer_than_upstream"]` and `info["version"]["outdated"]` seems to be related, aren't they?

cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
+        info["version"]["latest"]
       end
+
+    puts "#{formula_s} : #{current_s} ==> #{latest_s}"
+  end
+  private_class_method :print_latest_version
+
+  def self.checkable_urls(formula)
+    urls = []
+    urls << formula.head.url if formula.head
```
maxim-belkin, line 297: 1. Do we need this if the goal is to check for "stable" versions?
2. Shall this be protected by some flag, e.g. `--head`? 
2. Should this follow the stable urls instead of preceding them?

cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
-        "version" => {
-          "current"             => current.to_s,
-          "latest"              => latest.to_s,
-          "outdated"            => is_outdated,
-          "newer_than_upstream" => is_newer_than_upstream,
-        },
-      }
+    if Homebrew.args.debug?
+      puts "\nFormula:          #{formula_name(formula)}"
+      puts "Head only?:       #{!formula.stable?}" unless formula.stable?
```
maxim-belkin, line 348: HEAD only is checked with (from dev-cmd/audit.rb)
```
formula.head && formula.devel.nil? && formula.stable.nil?
```

cmd/livecheck.rb
```diff
@@ -114,13 +191,33 @@ def livecheck
...
     if formula.deprecated? && !formula.livecheckable?
       if Homebrew.args.json?
-        return {
-          "formula" => formula_name(formula),
-          "status"  => "deprecated",
-        }
+        return status_hash(formula, "deprecated")
       elsif !Homebrew.args.quiet?
         puts "#{Tty.red}#{formula_name(formula)}#{Tty.reset} : deprecated"
         return
```
maxim-belkin, line 223: If this return still required, provided that you factored out this part into a separate method?

cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
-          "formula" => formula_name(formula),
-          "status"  => "skipped",
-        }
-        json_hash["messages"] = [skip_msg] unless skip_msg.nil? || skip_msg.empty?
-        return json_hash
+        if skip_msg.nil? || skip_msg.empty?
+          return status_hash(formula, "skipped")
+        else
+          return status_hash(formula, "skipped", [skip_msg])
+        end
```
maxim-belkin, line 261: This suggestion is related to my comment about the `status_hash` method above. 
```suggestion
        return status_hash(formula, "skipped", [skip_msg])
```

cmd/livecheck.rb
samford: `newer_than_upstream == true` means that the formula version is newer than the `latest` version (i.e., the upstream source isn't giving us the newest version), so we display the formula version as red in the output.

`outdated == true` means the `latest` version is newer than the formula version, so we display the newer version as green in the output.

This behavior is very useful and I definitely wouldn't remove them simply to clean up the code in a very minor way (i.e., users see the output but they don't see the code). These are two separate concerns and the conditionals can't be combined.

The if/else blocks are used so that we're only setting the variable once (i.e., not setting `current_s` to `info["version"]["current"]` and then conditionally modifying it to include the coloring like `current_s = "#{Tty.red}#{current_s}#{Tty.reset}" if info["version"]["newer_than_upstream"]`). If anything changes here, I feel that the `formula_s` code should be updated to bring it in line with `current_s` and `latest_s` (avoiding the `+=` after the initial definition).

cmd/livecheck.rb
```diff
@@ -114,13 +191,33 @@ def livecheck
...
+    Homebrew.args.full_name? ? formula.full_name : formula
+  end
+  private_class_method :formula_name
+
+  def self.status_hash(formula, status_str, messages = nil)
+    status_hash = {
+      "formula" => formula_name(formula),
+      "status"  => status_str,
+    }
+    status_hash["messages"] = messages if messages.is_a?(Array)
```
maxim-belkin, line 204: Can this be made more general? E.g., use `if messages.present?` and then check if it (or its elements) `.respond_to? :str`? This would allow us to use `return status_hash(..., message)` without putting it into an `if/else` block only to not call this method incorrectly.

cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
+        # Truncate the URL at the user/repo part, if possible
+        github_repo_url = url[%r{((?:[a-z]+://)?github.com/[^/]+/[^/#]+)}]
+        url = github_repo_url unless github_repo_url.nil?
+
+        url = url[0..-2] if url.end_with?("/")
+
+        url += ".git"
+      end
+    elsif url.include?("/-/archive/")
+      url = url.sub(%r{/-/archive/.*$}i, ".git")
```
maxim-belkin, line 330: This block would be called for all urls even those that don't `github` in their name:

https://github.com/Homebrew/homebrew-livecheck/pull/1088/files#diff-350f6577a14157a255d00f694a342423R361

I'm not convinced we should be doing that

cmd/livecheck.rb
```diff
@@ -167,88 +254,178 @@ def print_latest_version(formula)
...
+    url
+  end
+  private_class_method :preprocess_url
 
-    formula_s = "#{Tty.blue}#{formula_name(formula)}#{Tty.reset}"
+  def self.latest_version(formula)
+    has_livecheckable = formula.livecheckable?
+    livecheck = formula.livecheck
+    livecheck_regex = livecheck.regex
+    livecheck_url = livecheck.url
```
maxim-belkin, line 341: Do we need these shortcuts? I'm not sure they are necessary.

cmd/livecheck.rb
samford: Changing the order that livecheck collects the formula URLs for the default check (i.e., when there's no `livecheck` block) would potentially affect a large number of formulae and this isn't a change that we can make lightly.

There's merit in revisiting this sometime in the future but it's far outside the scope of this PR. This is something that we would discuss in a dedicated issue/PR, where we can assess the functional differences of certain changes and establish benefits/detriments before making a decision.

cmd/livecheck.rb
samford: [`formula.devel` is deprecated](https://github.com/Homebrew/brew/pull/7688) and I don't see any usage in homebrew-core or linuxbrew-core.

With that in mind, I don't think it's unreasonable to treat that condition as basically `formula.head && formula.stable.nil?`. The main difference with livecheck is that we're not checking for the existence of `formula.head`, which I think is fine. That is to say, a formula must have either `stable` or `head`, so if `stable` isn't provided then `head` must be and that would be a "head-only" formula.

That said, I wasn't the one who implemented head-only formula checking (that was Dawid) and I was simply following the existing conditions he established.

cmd/livecheck.rb
maxim-belkin: Does newer_than_upstream mean >, and outdated < ?

cmd/livecheck.rb
maxim-belkin: What is the scope of this PR?

cmd/livecheck.rb
samford: These are the relevant definitions, where `current` is the formula version and `latest` is the "newest" upstream version:

```
"outdated"            => formula.stable? ? (current < latest) : (current != latest),
"newer_than_upstream" => formula.stable? && (current > latest)
```

For stable formulae, `outdated` means "formula version < newest upstream version" and `newer_than_upstream` means "formula version > newest upstream version".

cmd/livecheck.rb
maxim-belkin: Can formulae in taps have devel blocks?

cmd/livecheck.rb
samford: `skip_msg` defaults to an empty string (it's never `nil` despite that lingering `skip_msg.nil?` condition that we can probably remove), so there's the potential that the JSON output would include `"messages": [""]` unless we check for `skip_msg.empty?`.

The current JSON output only includes a message when it's not an empty string. I don't think there's value in an empty message string. There may be an argument for an empty messages array but that's better left for when we have the general discussion about the JSON output. For the purposes of this PR, I think it's reasonable to simply maintain the existing behavior.

cmd/livecheck.rb
samford: It sounds like you're suggesting allowing `status_hash(formula, "error", "Some error message")` alongside the existing `status_hash(formula, "error", ["Some error message"])`?

I considered allowing `messages` to be a string or an array of strings but I didn't feel that the additional complexity was worth it. Rather, anyone calling this method should simply follow the existing calls that use an array of strings as the `messages` argument.

If I've misinterpred this, please explain further.

cmd/livecheck.rb
samford: This condition relates to GitLab archive URLs (e.g., `https://gitlab.com/dps8m/dps8m/-/archive/R2.0/dps8m-R2.0.tar.gz`) not GitHub. Not all GitLab URLs are at gitlab.com (e.g., `code.videolan.org`, `git.openldap.org`, etc.), so we have to simply match any URL that contains `/-/archive/`. This isn't currently an issue and we'll address it if it ever does become an issue.

I explained this in #601 as follows:

> I renamed `github_preprocess()` to a more general name, `preprocess_url`, as it deals with more than just GitHub-related changes (e.g., the `/-/archive/` part is for GitLab URLs). This function is where we would make any changes in the URL before working with it, so it makes sense to use a general name.

cmd/livecheck.rb
samford: > What is the scope of this PR?

The changes that I outlined in the PR description. The only thing that changed about `checkable_urls()` in this PR was that it moved into the `Homebrew` module and became private. The code inside of the function is identical to the previous copy in `utils.rb`, so discussing functional changes to it isn't appropriate for this PR (especially in light of the time constraints). Any review should be specifically related to the changes that I've made here or new code that was introduced.

cmd/livecheck.rb
samford: `devel` is only deprecated at this point, so the related code is still in `Homebrew/brew` but it comes with a warning when the formula is instantiated:

```
Warning: Calling 'devel' blocks in formulae is deprecated! Use 'head' blocks or @-versioned formulae instead.
Please report this issue to the example/test tap (not Homebrew/brew or Homebrew/core), or even better, submit a PR to fix it:
  /usr/local/Homebrew/Library/Taps/example/homebrew-test/Formula/test.rb:7
```

A tap can have a `devel` block but I don't personally have any interest in supporting it, since it's on its way out.

That said, this is pre-existing behavior that wasn't changed in this PR, so we shouldn't be discussing it here. If you want to support `devel` blocks, open an issue or PR for it in Homebrew/brew after that migration is done.

cmd/livecheck.rb
maxim-belkin: My concern is that you're assuming that if site URL matches `/-/archive/`, then we're dealing with a Git repository. While possible, I'm not convinced this is true everywhere, as there is other DVCS-es, such as Mercurial, etc, and they're not restricted from using`/-/archive/` in their URLs.

samford: Sure, I'm not arguing that this is infallible or that there isn't a better way to approach this.

If you have an idea on how to improve this then we can discuss it in the future but it's not within the scope of this PR. The is pre-existing behavior and this PR didn't introduce or modify it.

cmd/livecheck.rb
maxim-belkin: Ah, got it. I didn't know it will be deprecated for taps as well. I thought this deprecation was homebrew/core-specific.

cmd/livecheck.rb
maxim-belkin: This line determines how `status_hash` method is called on lines 257--261. I find code where executive section is linear (such as do X, Y Z) to be easier to understand. I think it's OK to increase the complexity here a bit to handle various types of input (String and Array).

samford: Allowing the `messages` argument to be a string  (or an array of strings) merely allows you to avoid encapsulating the argument in an array. It doesn't remove the need to check `skip_msg.empty?`.

It's necessary to make sure that `skip_msg` isn't an empty string, otherwise `[skip_msg]` would be `[""]` and we don't want this as the `messages` variable. Your suggestion above could lead to this situation whereas the existing setup simply omits `messages` in that case.

If you're looking to get rid of the if/else setup and condense these lines, we could technically do the following, though it's a bit more difficult to read:

```ruby
return status_hash(formula, "skipped", (skip_msg.empty? ? nil : [skip_msg]))
```

Or `#status_hash` could be modified to deal with undesirable `messages` input:

```ruby
if messages.is_a?(Array)
  messages.filter! { |message| message.is_a?(String) && !message.empty? }
  status_hash["messages"] = messages unless messages.empty?
end
```

This would technically allow you to use `return status_hash(formula, "skipped", [skip_msg])` and `status_hash` would simply ignore the `[""]` array. Perhaps this is what you were describing?

---

I can potentially understand something like `return status_hash(formula, "skipped", (skip_msg.empty? ? nil : [skip_msg]))` but I don't think we should ever be passing `[""]` into `#status_hash` as the `messages` argument (as in `return status_hash(formula, "skipped", [skip_msg])`).

I'm of the opinion that since `#status_hash` is a private method, anyone who calls it within livecheck should simply make sure to A) pass an array of strings (none of which are empty) or `nil` or B) omit `messages` from the `#status_hash` call. If this were a public method, then I can understand adding the aforementioned logic to address and filter out bad input.

With this in mind, I pushed a commit that condenses the `skip_msg` lines, which I think is your main goal with this discussion.

cmd/livecheck.rb
samford: I pushed a commit that uses `return status_hash(formula, "skipped", (skip_msg.empty? ? nil : [skip_msg]))` instead (as it addresses my issue with using `[skip_msg]` when it could be `[""]`). It's a little more difficult to read but does this do it for you?

cmd/livecheck.rb
samford: [`formula#livecheck` is a method that returns the @livecheck object](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L2712-L2718), so `formula.livecheck.url` involves two method calls: one to return the `livecheck` object and another to get the `url`. The `livecheck` object methods are found here: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/livecheck.rb

It's technically possible to just use `formula.livecheck.url` but it would end up being more expensive throughout `#latest_version`, which is why we simply store the values up front and use those throughout `latest_version` instead.

samford:
Thanks for the review. It may be that this takes a different shape during the Homebrew/brew migration but I think this is more likely to be correct than the `heuristic.rb` and `utils.rb` setup, at least. There's plenty of room for improvement in livecheck but this is one step forward (with more to come).

PR #1092, opened 2020-07-12
python: rename to python@3.8, update regex

nandahkrishna:
Renaming `python.rb` to `python@3.8.rb` as the Formula in core is `python@3.8`. Prior to this change, livecheck would give `python@3.8 : versioned` as the output for `brew livecheck python`.

### DIFF
--- a/Livecheckables/python@3.8.rb
+++ b/Livecheckables/python@3.8.rb
@@ -1,6 +1,6 @@
 class PythonAT38
   livecheck do
-    url "https://www.python.org/downloads/"
-    regex(%r{href=.*?ftp/python/(\d+(?:\.\d+)+)/}i)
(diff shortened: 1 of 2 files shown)


Livecheckables/python@3.8.rb
```diff
@@ -0,0 +1,6 @@
+class PythonAT38
+  livecheck do
+    url "https://www.python.org/downloads/"
+    regex(%r{href=.*?ftp/python/(\d+(?:\.\d+)+)/}i)
```
samford, line 4:
```suggestion
    url "https://www.python.org/ftp/python/"
    regex(%r{href=.*?v?(3\.8(?:\.\d+)*)/?["' >]}i)
```

PR #1116, opened 2020-07-13
jhiccup: update regex

samford:
This brings the existing `jhiccup` livecheckable up to current standards:

* Use `href=.*?` (instead of `href=".*?/`, in this case)
* Replace the delimiter between software name and version in file name with `[._-]`
* Use `v?(\d+(?:\.\d+)+)` instead of `([0-9.]+)`
* Make regex case insensitive unless case sensitivity is needed

### DIFF
--- a/Livecheckables/jhiccup.rb
+++ b/Livecheckables/jhiccup.rb
@@ -1,6 +1,6 @@
 class Jhiccup
   livecheck do
     url :homepage
-    regex(%r{href=".*?/jHiccup-([0-9.]+)-dist})
+    regex(/href=.*?jHiccup[._-]v?(\d+(?:\.\d+)+)-dist\.zip/i)
   end
 end


nandahkrishna:
I'm not sure but I recall changing the delimiter between the version and `dist` to `[._-]`. I think it was `musepack` and `wolfssl`.

samford:
> I'm not sure but I recall changing the delimiter between the version and `dist` to `[._-]`

I considered it and I've done it in other formulae but I wasn't entirely sure if it would be worth it here. Right now we're in a weird in between space where we use explicit hyphens in some places and `[._-]` in others (e.g., `some-software-name[._-]`). In my mind, allowing `-dist` is the same as allowing hyphens in the software name before the `[._-]` separator.

There's an argument for using `[._-]` instead but it's late and my brain is tired, so I'm going to keep punting this decision.

PR #1125, opened 2020-07-13
Improve regexes containing `href=.+` and `href=.+?`

nandahkrishna:
This PR improves regexes containing `href=.*` in the following ways:
* `href=.+` or `href=.+?` –> `href=.*?`
* Delimiter between software name and version string, `-` –> `[._-]`
* The regexes are now case-insensitive
* Additional fixes in
    * `icoutils`
    * `scamper`
    * `szip`
    * `x264`

### DIFF
--- a/Livecheckables/szip.rb
+++ b/Livecheckables/szip.rb
@@ -1,6 +1,6 @@
 class Szip
   livecheck do
     url "https://support.hdfgroup.org/ftp/lib-external/szip/"
-    regex(%r{href=.+?v?(\d+(?:\.\d+)+)/?["']})
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?["' >]}i)
   end
 end
--- a/Livecheckables/x264.rb
+++ b/Livecheckables/x264.rb
@@ -4,6 +4,6 @@ class X264
   # the version information at the time of writing.
   livecheck do
     url "https://artifacts.videolan.org/x264/release-macos/"
-    regex(/href=.+?x264-(r\d+)-/i)
+    regex(%r{href=.*?x264[._-](r\d+)[._-][a-z0-9]+/?["' >]}i)
   end
 end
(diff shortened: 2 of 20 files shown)


Livecheckables/szip.rb
```diff
@@ -1,6 +1,6 @@
 class Szip
   livecheck do
     url "https://support.hdfgroup.org/ftp/lib-external/szip/"
-    regex(%r{href=.+?v?(\d+(?:\.\d+)+)/?["']})
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?[ '">]}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?["' >]}i)
```

The order that I've been using is `["' >]` and this is used in a number of livecheckables right now. There do appear to be a handful of uses of `[ '">]` (11 in total) but we'll need to bring them up to date sometime afterward.

Livecheckables/x264.rb
```diff
@@ -4,6 +4,6 @@ class X264
   # the version information at the time of writing.
   livecheck do
     url "https://artifacts.videolan.org/x264/release-macos/"
-    regex(/href=.+?x264-(r\d+)-/i)
+    regex(%r{href=.*?x264[._-](r\d+)[._-][\da-z]+/?[ '">]}i)
```
samford, line 7:
```suggestion
    regex(%r{href=.*?x264[._-](r\d+)[._-][a-z0-9]+/?["' >]}i)
```

`[a-z0-9]` may be a bit more readable/clear than `[\da-z]` within the context of the character class. As above, I've updated this to use `["' >]` (instead of the differing order).

nandahkrishna: There seem to be a few Livecheckables using `[\da-z]`, shall I add that to the cleanup list?

samford: Sure, there only appear to be a few, so it wouldn't take long.

PR #1127, opened 2020-07-13
Change `href=".*` and `href=".*?` to `href=.*?`

nandahkrishna:
This PR changes `href=".*` and `href=".*?` in regexes (when followed by filename) to `href=.*?`.


nandahkrishna:
Thanks @samford!

PR #1141, opened 2020-07-14
Change `href="` followed by name to `href=.*?`

nandahkrishna:
This PR changes `href="` when followed by the file/formula name to `href=.*?`.

The changes made are, for the most part, just simple replacements as stated above, except:
* `exempi`: also removed `[^"']`
* `icbirc`: also removed leading `/` and hence changing `%r{...}` to `/.../`
* `mkvtoolnix`: also removed `[^"]`

I've run `brew style homebrew/livecheck` and ensured that these changes haven't broken any Livecheckables.


nandahkrishna:
Ah, should've spotted that. Thanks!

PR #1144, opened 2020-07-14
bashdb: update regex

nandahkrishna:
This PR updates the regex for `bashdb`.

The changes are:
* `href="` to `href=.*?` and removed the path from the regex
* They've been using `x.x-x.x(.x)?` since 2004, the version capturing group has been updated accordingly (needs to be checked, of course).
* `v?` prior to version capturing group
* `/` matching changed to `/?["' >]`
* Case-insensitivity

### DIFF
--- a/Livecheckables/bashdb.rb
+++ b/Livecheckables/bashdb.rb
@@ -1,6 +1,6 @@
 class Bashdb
   livecheck do
     url "https://sourceforge.net/projects/bashdb/files/bashdb/"
-    regex(%r{href=.*?v?(\d+(?:\.\d+)+[._-](?:\d+(?:\.\d+)+))/?["' >]}i)
+    regex(%r{href=(?:["']|.*?bashdb/)?v?(\d+(?:[.-]\d+)+)/?["' >]}i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/bashdb.rb
```diff
@@ -1,6 +1,6 @@
 class Bashdb
   livecheck do
     url "https://sourceforge.net/projects/bashdb/files/bashdb/"
-    regex(%r{href="/projects/bashdb/files/bashdb/([0-9.\-]+)/"})
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+[._-](?:\d+(?:\.\d+)+))/?["' >]}i)
```
samford, line 4:
```suggestion
    regex(%r{href=(?:["']|.*?bashdb/)?v?(\d+(?:[._-]\d+)+)/?["' >]}i)
```

nandahkrishna: Do we want the delimiter matching portion to be `[._-]` or `[.-]`? I thought we might go with the latter for the sake of consistency. Also, they haven't used `_` anywhere.

samford: Within the context of this formula, `[.-]` is fine in `v?(\d+(?:[.-]\d+)+)`. I was using `[._-]` before because we do have some formulae that use underscores to separate the numeric parts of the version upstream (e.g., `atari800` uses `4_2_0` for `4.2.0`, `nsd` uses `4_3_1` for `4.3.1`, etc.).

I updated the suggestion with this in mind.

PR #1145, opened 2020-07-14
libwps: update regex

nandahkrishna:
This PR updates `libwps`'s regex.

Changes:
* `href="` –> `href=.*?` and removal of leading path
* `/` matching –> `/?["' >]`
* `-` –> `[._-]` (delimiter between formula name and version)
* `v?` before version capturing group
* Version capturing group updated to current standard
* Case-insensitivity

### DIFF
--- a/Livecheckables/libwps.rb
+++ b/Livecheckables/libwps.rb
@@ -1,6 +1,6 @@
 class Libwps
   livecheck do
     url "https://sourceforge.net/projects/libwps/files/libwps/"
-    regex(%r{href=.*?libwps[._-]v?(\d+(?:\.\d+)+)/?["' >]}i)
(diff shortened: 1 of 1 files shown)


Livecheckables/libwps.rb
```diff
@@ -1,6 +1,6 @@
 class Libwps
   livecheck do
     url "https://sourceforge.net/projects/libwps/files/libwps/"
-    regex(%r{href="/projects/libwps/files/libwps/libwps-([0-9.]+)/"})
+    regex(%r{href=.*?libwps[._-]v?(\d+(?:\.\d+)+)/?["' >]}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?libwps(?:/|[._-])v?(\d+(?:\.\d+)+)/?["' >]}i)
```

PR #1150, opened 2020-07-15
Add Xorg strategy

samford:
Some of the X.Org formulae from linuxbrew/xorg are being merged into homebrew-core in Homebrew/homebrew-core#57995. With this in mind, livecheck needs to handle the checks for these formulae. The linuxbrew/xorg tap has an `Xorg` strategy (created by Maxim) that's used on these formulae, so I'm incorporating it into livecheck as a built-in strategy.

The `Xorg` strategy wasn't quite working properly for all of the cases that I encountered, so I made some improvements:

* Modified the `match?` logic to only match URLs the strategy can use (we can expand this as needed)
* Reworked the way that the `package_name`, `page_url`, and `regex` are identified and generated, so it should be a little more reliable than the previous method
* Added support for caching page content, so we don't end up fetching the same index pages over and over (this also significantly speeds up the `Xorg` strategy for runs with more than one formula)

This works with all the applicable formulae in the linuxbrew/xorg tap, though we'll need to remove some broken `livecheck` blocks in those formulae to allow the `Xorg` strategy to properly work. I'll be updating these formulae after this PR is merged, as well as removing the `Xorg` strategy from that tap (so this one is used instead).

### DIFF
--- a/livecheck/livecheck_strategy/xorg.rb
+++ b/livecheck/livecheck_strategy/xorg.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require "open-uri"
+
+module LivecheckStrategy
+  class Xorg
+    NAME = name.demodulize
+    NICE_NAME = "X.Org"
+
+    @page_data = {}
+
+    def self.match?(url)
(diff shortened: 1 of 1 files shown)


iMichka:
Thanks !

PR #1152, opened 2020-07-15
Change `href=".*?/` to `href=.*?`

nandahkrishna:
This PR changes `href=".*?/` to `href=.*?`. It turns out the `/` (leading) wasn't required in any of the regexes, which I verified by comparing pre- and post-change debug outputs.

### DIFF
--- a/Livecheckables/aircrack-ng.rb
+++ b/Livecheckables/aircrack-ng.rb
@@ -1,6 +1,6 @@
 class AircrackNg
   livecheck do
     url :homepage
-    regex(%r{href=".*?/aircrack-ng-([0-9.]+)\.t})
+    regex(/href=.*?aircrack-ng-([0-9.]+)\.t/)
   end
 end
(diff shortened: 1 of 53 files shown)


nandahkrishna:
The only 'exception' here is `pike`, where the entire prefix `/latest-stable/` was unnecessary (only the stable version is linked to on the homepage).

nandahkrishna:
In the second commit (please excuse the message, didn't know what'd be ideal), I've made all the suggested changes and also converted the PyPI urls of those Livecheckables modified in this PR to `:stable`.

However `dq` fails because of a redirection error. Any idea on how to handle it @samford?

samford:
> However qd fails because of a redirection error. Any idea on how to handle it @samford?

It sounds like you may need to run `brew update` to get the latest `qd` formula. I updated the URLs yesterday to avoid the previous redirections, so if you haven't updated since then you'll receive a redirection error when using `url :homepage`.

nandahkrishna:
Haha, just realised it myself 😅

PR #1155, opened 2020-07-15
chkrootkit: update regex

nandahkrishna:
This PR updates `chkrootkit`'s regex to conform to current standards. I took inspiration from `pidof` for this one, but there may be better alternatives.

### DIFF
--- a/Livecheckables/chkrootkit.rb
+++ b/Livecheckables/chkrootkit.rb
@@ -1,6 +1,6 @@
 class Chkrootkit
   livecheck do
     url :homepage
-    regex(/href=.*?download[^>]+>chkrootkit v?(\d+(?:\.\d+)+)/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/chkrootkit.rb
```diff
@@ -1,6 +1,6 @@
 class Chkrootkit
   livecheck do
     url :homepage
-    regex(%r{href="/download/">chkrootkit ([0-9,.]+)})
+    regex(/href=.*?download[^>]+>chkrootkit v?(\d+(?:\.\d+)+)/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?download[^>]*>chkrootkit v?(\d+(?:\.\d+)+)/i)
```

Making this optional would allow for `href=download>`, which may happen if an upstream website doesn't use `"` or `'` around the `href` attribute (which I've unfortunately seen before).

PR #1161, opened 2020-07-15
fs-uae: update url and regex

nandahkrishna:
This PR updates `fs-uae`'s `url` to that of an index page of all releases, and also brings the `regex` up to current standards.

### DIFF
--- a/Livecheckables/fs-uae.rb
+++ b/Livecheckables/fs-uae.rb
@@ -1,6 +1,6 @@
 class FsUae
   livecheck do
-    url "http://fs-uae.net/stable/"
-    regex(%r{href=.*?v?(\d+(?:\.\d+)+(?:u\d+)?)/?["' >]}i)
+    url "https://fs-uae.net/download"
+    regex(/href=.*?fs-uae[._-]v?(\d+(?:\.\d+)+)\.t/i)
   end
 end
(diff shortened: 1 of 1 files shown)


Livecheckables/fs-uae.rb
```diff
@@ -1,6 +1,6 @@
 class FsUae
   livecheck do
-    url :homepage
-    regex(%r{Latest Version.*?href="/download">([0-9.]+)<}m)
+    url "http://fs-uae.net/stable/"
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+(?:u\d+)?)/?["' >]}i)
```
samford, line 4:
```suggestion
    url "https://fs-uae.net/download"
    regex(/href=.*?fs-uae[._-]v?(\d+(?:\.\d+)+)\.t/i)
```

nandahkrishna:
That makes sense, thanks for the explanation! I'll make sure to keep that in mind for future PRs.

samford:
Just to be clear, we'll continue favoring directory listing pages when they're closest to the stable archive and the archive files are available on the page.

That is to say, this specifically applies to the situation where we have a directory listing page with directories like `1.2.3/` and we have an alternative first-party page that provides a link to the archive file. Similarly, if we only have the directory listing page (no alternative first-party pages or alternative sources) and all we have are directories like `1.2.3/`, we'll have to use that.

That may have been clear above but I just wanted to make sure. Of course, let me know if this still isn't clear or if you have any questions.

nandahkrishna:
That was clear, yes. Thanks again!

PR #1163, opened 2020-07-15
ghc: update url and regex

nandahkrishna:
This PR updates `ghc`'s `url` to that of an index page of all releases (which aligns with the stable archive location). The `regex` has also been updated as a result, and it conforms to current standards.

### DIFF
--- a/Livecheckables/ghc.rb
+++ b/Livecheckables/ghc.rb
@@ -1,6 +1,6 @@
 class Ghc
   livecheck do
-    url "https://www.haskell.org/ghc/download.html"
-    regex(/href="download_ghc_(?:\d+(?:_?\d+)+).html">(\d+(?:\.\d+)+)</)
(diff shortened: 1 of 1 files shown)


Livecheckables/ghc.rb
```diff
@@ -1,6 +1,6 @@
 class Ghc
   livecheck do
-    url "https://www.haskell.org/ghc/download.html"
-    regex(/href="download_ghc_(?:\d+(?:_?\d+)+).html">(\d+(?:\.\d+)+)</)
+    url "https://downloads.haskell.org/~ghc/"
+    regex(%r{href=.*?v?(\d+(?:\.\d+)+)/?["' >]}i)
```
samford, line 4:
```suggestion
    regex(%r{href=.*?v?(\d+(?:\.\d{1,4})+)/?["' >]}i)
```

PR #1172, opened 2020-07-16
r: update regex

nandahkrishna:
This PR updates `r`'s `regex` to conform to current standards.

I checked if we could change the url to align with that of the stable archive, but the files are organised into directories and the filename of the latest version is `R-latest`, so I've not changed the url.

### DIFF
--- a/Livecheckables/r.rb
+++ b/Livecheckables/r.rb
@@ -1,6 +1,6 @@
 class R
   livecheck do
     url "https://cran.rstudio.com/banner.shtml"
-    regex(%r{href="src/base/R-.*>R-([\d.]+)\.t})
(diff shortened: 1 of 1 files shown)


Livecheckables/r.rb
```diff
@@ -1,6 +1,6 @@
 class R
   livecheck do
     url "https://cran.rstudio.com/banner.shtml"
-    regex(%r{href="src/base/R-.*>R-([\d.]+)\.t})
+    regex(/href=.*?R[._-]v?(\d+(?:\.\d+)+)\.t/i)
```
samford, line 4:
```suggestion
    regex(%r{href=(?:["']?|.*?/)R[._-]v?(\d+(?:\.\d+)+)\.t}i)
```

PR #1189, opened 2020-07-19
Make Livecheckable regexes case-insensitive

nandahkrishna:
This PR makes Livecheckable regexes case-insensitive. I believe nearly all Livecheckables will have case-insensitive regexes, except `boost`, if I remember correctly. It requires updation anyway so that can be handled separately.

I compared the livecheck output before and after the changes a couple of times, but it would be great if you could have a look at it as well @samford, to ensure everything is working as intended. Thanks!


nandahkrishna:
Thanks!

PR #1203, opened 2020-07-22
Update ([0-9.]+) in regexes

nandahkrishna:
Sorry for the delay with this PR and the less descriptive commit message.

In this PR, I've updated those Livecheckables with `([0-9.]+)` regexes, where the version we're trying to match is neither a year/timestamp nor just a natural number (as is the case with `txr`), to `(\d+(?:\.\d+)+)`.

I haven't, however, prepended the `v?` in this PR. Should I do that as well here? Thanks.

### DIFF
--- a/Livecheckables/abnfgen.rb
+++ b/Livecheckables/abnfgen.rb
@@ -1,6 +1,6 @@
 class Abnfgen
   livecheck do
     url :homepage
-    regex(%r{href=.*?/abnfgen-([0-9.]+)\.t}i)
+    regex(%r{href=.*?/abnfgen-v?(\d+(?:\.\d+)+)\.t}i)
   end
 end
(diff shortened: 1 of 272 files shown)


samford:
> I haven't, however, prepended the `v?` in this PR. Should I do that as well here?

Yes, please. The idea is that eventually we may be able to provide some built-in snippets in place of standardized parts of regex (e.g., `href=.*?`, `[._-]`, `v?(\d+(?:\.\d+)+)`, etc.) and this helps with that.

nandahkrishna:
I've made sure to check whether `v?` was required or not (for example, `autogen` does not, as it uses `rel`).

Unrelated to this, I found out while going through the Livecheckables now that we have some `HREF+"` (all caps) that we missed changing. I could make those changes in a separate PR if you want me to.

PR #1209, opened 2020-07-23
Remove LivecheckStrategy NAME requirement

samford:
Currently `LivecheckStrategy` classes are required to include `NAME = name.demodulize.freeze`, so that we avoid running this code numerous times within `#latest_version`. Ideally this should be handled within livecheck, so it's not necessary to include this snippet in all `LivecheckStrategy` classes.

This PR aims to achieve this by creating and populating a `@livecheck_strategy_names` hash, which maps a `LivecheckStrategy` to the demodulized strategy name. With this setup, we continue to only run `strategy.name.demodulize` once per strategy over a given run of `brew livecheck`.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -42,6 +42,8 @@ module Homebrew
   HEAD_ONLY_MSG = "HEAD only formula must be installed to be livecheckable"
   private_constant :NO_VERSIONS_MSG, :HEAD_ONLY_MSG
 
+  @livecheck_strategy_names = {}
+
   def livecheck_args
     Homebrew::CLI::Parser.new do
       usage_banner <<~EOS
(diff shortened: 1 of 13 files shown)


cmd/livecheck.rb
```diff
@@ -122,6 +124,12 @@ def livecheck
       Dir[File.join(tap_strategy_path, "*.rb")].sort.each(&method(:require)) if Dir.exist?(tap_strategy_path)
     end
 
+    LivecheckStrategy.constants.sort.each do |strategy_symbol|
```
maxim-belkin, line 127: I think it would be nice to add a comment (above this line) explaining why we built a hash of all strategy names instead of using `.each(&:...` approach.

samford: I added `# Cache demodulized strategy names, to avoid repeating this work` above this line. What do you think about this?

maxim-belkin: It's definitely better. 👍 You could add something like "when processing multiple formulae/taps" but this is up to you.

samford: I feel like the conditions where it avoids extra work are too lengthy to explain in a brief comment, so I think I may just leave it at this. Anyone who's interested can look through the code to see where/how `@livecheck_strategy_names` is used.

That is, this saves work when:

1. Running `brew livecheck` with multiple formulae
2. Running `brew livecheck` on a formula that iterates through multiple URLs (we want to avoid this in the future but it's currently true for some formulae)
3. Running `brew livecheck --debug`
4. Running `brew livecheck --json --verbose`

PR #1210, opened 2020-07-23
Update HREF=" to href=.*? in regexes

nandahkrishna:
This PR updates instances of `HREF="` to `href=.*?` in Livecheckable regexes. I believe CI might fail (specifically for `hspell`) as SourceForge is experiencing problems right now.

Update (as expected):
```
Error: hspell: 503 Service Temporarily Unavailable
```

### DIFF
--- a/Livecheckables/dwarfutils.rb
+++ b/Livecheckables/dwarfutils.rb
@@ -1,6 +1,6 @@
 class Dwarfutils
   livecheck do
     url :homepage
-    regex(/href=.*?libdwarf-([0-9.]+)\.t/i)
+    regex(%r{href=(?:["']?|.*?/)libdwarf-([0-9.]+)\.t}i)
   end
 end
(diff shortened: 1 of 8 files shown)


Livecheckables/dwarfutils.rb
```diff
@@ -1,6 +1,6 @@
 class Dwarfutils
   livecheck do
     url :homepage
-    regex(/HREF="libdwarf-([0-9.]+)\.t/i)
+    regex(/href=.*?libdwarf-([0-9.]+)\.t/i)
```
samford, line 4:
```suggestion
    regex(%r{href=(?:["']?|.*?/)libdwarf-([0-9.]+)\.t}i)
```

samford, line 4:
```suggestion
    regex(%r{href=(?:["']?|.*?/)libdwarf-([0-9.]+)\.t}i)
```

There's a link to an alpha release with the format `alpha-libdwarf-20180706.tar.gz` that we should avoid. The relevant HTML looks like this:

```html
<A HREF="alpha-libdwarf-20180706.tar.gz" id="alpha-libdwarf-20180706">alpha-libdwarf-20180706.tar.gz</A>
```

This suggestion uses the typical snippet for `href` attributes that may only include the file name.

PR #1211, opened 2020-07-23
Update ([0-9.]+) to v?(\d{6,8}) in regexes

nandahkrishna:
This PR depends on #1210 due to changes in `dwarfutils`.

I've updated `([0-9.]+)` in date-based version regexes to `v?(\d{6,8})`.

CI may fail for `wtf`'s livecheck as SourceForge is having problems.

### DIFF
--- a/Livecheckables/acpica.rb
+++ b/Livecheckables/acpica.rb
@@ -1,6 +1,6 @@
 class Acpica
   livecheck do
     url "https://acpica.org/downloads"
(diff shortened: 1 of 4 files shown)


samford:
I tested this with the requested changes for `dwarfutils` in #1210 and it looks good.

As we discussed, once the requested change in #1210 is integrated and I've merged that PR, this can be rebased and it should be good to go.

PR #1217, opened 2020-07-26
abcmidi: update regex

nandahkrishna:
This PR brings `abcmidi`'s regex up to current standards. They seem to be using a YYYY.MM.DD format, however I've been strict only with `\d{4}` for the year part of the version. Let me know if there are any changes, thanks!

### DIFF
--- a/Livecheckables/abcmidi.rb
+++ b/Livecheckables/abcmidi.rb
@@ -1,6 +1,6 @@
 class Abcmidi
   livecheck do
     url :homepage
-    regex(/abcMIDI-(20[0-9]+\.[0-9]+\.[0-9]+)/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/abcmidi.rb
```diff
@@ -1,6 +1,6 @@
 class Abcmidi
   livecheck do
     url :homepage
-    regex(/abcMIDI-(20[0-9]+\.[0-9]+\.[0-9]+)/i)
+    regex(/abcMIDI[._-]v?(\d{4}(?:\.\d+)+)\.zip/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?abcMIDI[._-]v?(\d{4}(?:\.\d+)+)\.zip/i)
```

PR #1218, opened 2020-07-27
cppad: update regex

nandahkrishna:
This PR brings `cppad`'s regex up to current standards. They've consistently used a scheme that's 8 digits followed. by a dot and then a digit (could be digits). I didn't know what'd be the best solution here, but I've gone ahead with `(\d{8}(?:\.\d+)+)`, let me know if I should use something else.

### DIFF
--- a/Livecheckables/cppad.rb
+++ b/Livecheckables/cppad.rb
@@ -1,6 +1,6 @@
 class Cppad
   livecheck do
     url :head
-    regex(/(20[0-9]+\.[0-9]+)/i)
+    regex(/^v?(\d{8}(?:\.\d+)+)$/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/cppad.rb
```diff
@@ -1,6 +1,6 @@
 class Cppad
   livecheck do
     url :head
-    regex(/(20[0-9]+\.[0-9]+)/i)
+    regex(/^v?(\d{8}(?:\.\d+)+)$/i)
```
samford, line 4:
```suggestion
    regex(/^v?(\d+(?:\.\d+)+)$/i)
```

PR #1223, opened 2020-07-27
Add strategy to livecheck DSL

### LINKED ISSUE #301: Add `using` option to explicitly specify the strategy to use

samford:
This is an implementation of the oft-mentioned `strategy` feature (née `using`). As described in #301, the purpose of this is to specify the strategy to use in a livecheckable. This is the end result of a number of internal changes that enabled this feature to exist.

Specifying the `strategy` in a livecheckable is generally only necessary when the default strategy that livecheck selects isn't the one that we want. For example, the `Sourceforge` strategy may apply to a given URL but we want to use `PageMatch` instead.

An example of this is as follows:

```ruby
livecheck do
  url :stable
  strategy :page_match
  regex(/href=.*?example[._-]v?(\d+(?:\.\d+)+)\.t/i)
end
```

The argument for `strategy` is a symbol corresponding to the strategy name (with an underscore between words, as in the strategy file names).

Though I initially described some additional functionality for this feature, I think this simple implementation is all that we'll need for now. If we need additional features, we can always expand it in the future.

The only other thing to note is that it was necessary to extend Homebrew/brew's `livecheck.rb` file to implement `strategy` in the `livecheck` DSL. This will be integrated back into Homebrew/brew when that migration occurs.

Closes #301.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -4,6 +4,7 @@ require "cli/parser"
 
 require_relative "../livecheck/livecheck_strategy"
 require_relative "../livecheck/extend/formulary"
+require_relative "../livecheck/extend/livecheck"
 
 module Homebrew
   module_function
@@ -350,6 +351,7 @@ module Homebrew
     has_livecheckable = formula.livecheckable?
     livecheck = formula.livecheck
     livecheck_regex = livecheck.regex
+    livecheck_strategy = livecheck.strategy
     livecheck_url = livecheck.url
 
     urls = [livecheck_url] if livecheck_url.is_a?(String) && !livecheck_url.blank?
@@ -370,9 +372,11 @@ module Homebrew
         next
       end
 
-      url = preprocess_url(original_url)
+      # Do not preprocess the URL when livecheck.strategy is set to :page_match
+      url = (livecheck_strategy == :page_match) ? original_url : preprocess_url(original_url)
+
       strategies = LivecheckStrategy.from_url(url, livecheck_regex.present?)
-      strategy = strategies[0]
+      strategy = LivecheckStrategy.from_symbol(livecheck_strategy) || strategies[0]
       strategy_name = @livecheck_strategy_names[strategy]
 
       if Homebrew.args.debug?
@@ -382,6 +386,16 @@ module Homebrew
         end
         puts "Strategy:         #{strategy.nil? ? "None" : strategy_name}"
         puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
+      end
+
+      if livecheck_strategy == :page_match && !livecheck_regex.present?
+        odebug "#{strategy_name} strategy requires a regex"
+        next
+      end
+
+      if livecheck_strategy.present? && !strategies.include?(strategy)
+        odebug "#{strategy_name} strategy does not apply to this URL"
+        next
       end
 
       next if strategy.nil?
--- a/livecheck/extend/livecheck.rb
+++ b/livecheck/extend/livecheck.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# Livecheck can be used to check for newer versions of the software.
+# The livecheck DSL specified in the formula is evaluated the methods
+# of this class, which set the instance variables accordingly. The
+# information is used by brew livecheck when checking for newer versions
+# of the software.
+class Livecheck
+  # The reason for skipping livecheck for the formula.
+  # e.g. `Not maintained`
+  attr_reader :skip_msg
+
+  def initialize(formula)
+    @formula = formula
+    @regex = nil
+    @skip = false
+    @skip_msg = nil
+    @strategy = nil
+    @url = nil
+  end
+
+  # Sets the strategy instance variable to the argument given or returns the
+  # strategy instance variable when no argument is given.
+  # @param symbol [Symbol] corresponding symbol for desired LivecheckStrategy
+  # @return [Symbol, nil]
+  def strategy(symbol = nil)
+    return @strategy if symbol.nil?
+
+    @strategy = symbol if symbol.is_a?(Symbol)
+  end
+
+  # Returns a Hash of all instance variable values.
+  def to_hash
+    {
+      "regex"    => @regex,
+      "skip"     => @skip,
+      "skip_msg" => @skip_msg,
+      "strategy" => @strategy,
+      "url"      => @url,
+    }
+  end
+end


cmd/livecheck.rb
```diff
@@ -361,9 +363,15 @@ def self.latest_version(formula)
...
+      # Do not process the URL when livecheck.strategy is set to :page_match
+      url = if livecheck.strategy && livecheck_strategy == LivecheckStrategy::PageMatch
+        original_url
+      else
+        preprocess_url(original_url)
+      end
+
       strategies = LivecheckStrategy.from_url(url, livecheck_regex.present?)
-      strategy = strategies[0]
+      strategy = livecheck_strategy || strategies[0]
```
maxim-belkin, line 374: What if `strategies` is empty?

cmd/livecheck.rb
```diff
@@ -374,6 +382,17 @@ def self.latest_version(formula)
         puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
       end
 
+      if livecheck_strategy == LivecheckStrategy::PageMatch &&
+         !livecheck_regex.present?
+        odebug "#{livecheck_strategy::NAME} strategy requires a regex"
```
maxim-belkin, line 387: will `::NAME` be valid after #1209?

cmd/livecheck.rb
samford: If `strategies` is empty, `strategies[0]` gives `nil`.

cmd/livecheck.rb
samford: This PR reflects the current context before #1209 and I'm planning to rebase/update it if #1209 is merged before it (or vice versa). I appreciate the reminder.

samford: I've updated this now that #1209 is merged.

cmd/livecheck.rb
```diff
@@ -384,6 +392,17 @@ def self.latest_version(formula)
...
       end
 
+      if livecheck_strategy == LivecheckStrategy::PageMatch &&
+         !livecheck_regex.present?
+        odebug "#{@livecheck_strategy_names[livecheck_strategy]} strategy requires a regex"
+        next
+      end
+
+      unless livecheck_strategy.nil? || strategies.include?(livecheck_strategy)
+        odebug "#{@livecheck_strategy_names[livecheck_strategy]} strategy does not apply to this URL"
```
maxim-belkin, line 402: I wonder if this (+ the one above) `odebug` should be `odie` instead.

samford: `odie` would break multi-formula runs. That is, the run would halt at `odie` and subsequent formulae wouldn't be checked.

maxim-belkin: True and I think this is what we want (fail if livecheck block specifies incorrect strategy).

samford: `odie` isn't the behavior we want in a multi-formula run, as it would prevent livecheck from checking any subsequent formulae.

For example:

* We have formulae A, B, C, and D
* B encounters a Git error
* `brew livecheck a b c d` would check A and `odie` on B, so C and D wouldn't be checked

Any issue with B shouldn't affect the run overall or prevent subsequent formulae from being checked.

maxim-belkin: I'm not sure if `livecheck` is more like `install` or `style`... I think of it as of `install` that doesn't attempt to continue if there is a problem.

samford: Checks aren't dependent upon one another, so there's no reason why one failing should cause execution to halt. This would be a huge pain when doing a long `--json` run, as the JSON output isn't printed until the very end, so all of the work would be thrown away if one formula fails (and for a completely avoidable and arbitrary reason). I can't see any benefit in using `odie` here whereas there are very clear detriments.

maxim-belkin: I don't see any benefit in such behavior either, but... again, `brew install f1 f2` wouldn't install `f2` if Homebrew failed to install `f1`. I thought that `style`, `audit` and `tests` are the only commands that are allowed to continue past failures. It might be that `livecheck` will join this pool but I wanted to bring this up to your attention.

samford: `livecheck` already continues past errors for a multi-formula run and this isn't anything new. The concerns of `brew install` are very different and there's currently no reason why `livecheck` should halt a run when one formula can't be checked properly. It would be tantamount to `brew style` halting on the first file that it encounters with issues, which would make it a pain to style-check multiple formulae in one run (e.g., `brew style homebrew/livecheck`).

Halting or pausing execution is something that we should be working to avoid when we encounter it, as it worsens the user experience. It's not something that we want to be deliberately introducing, as with `odie`.

maxim-belkin: Well, I explained my concern to the best of my abilities. I agree with what you say about impoving user experience but I'm not sure this `odebug` improves it. From what I understand, 
```
if brew livecheck x; then echo "success"; else echo "failed"; fi
```
 would never print "failed" and a user/CI would have hard time detecting an issue using this approach.

samford: `brew livecheck x` would still fail with the usual `Unable to get versions` error in this context (unless a subsequent URL works). As such, `failed` prints for a check that includes a `url` and either A) `strategy :page_match` but no `regex` or B) `strategy` with a strategy that doesn't apply to the URL.

The `odebug` statements are simply additional debug output and the important detail here is that it doesn't affect the existing success/failure conditions outside of this PR. As before, if a check fails, the user can simply run `brew livecheck --debug x` to understand why the check isn't working properly.

samford:
Understandably but they're each different things:

1. `livecheck.strategy`: This is the `strategy` symbol from the `livecheck` block.
2. `livecheck_strategy`: This is the strategy that corresponds to the `strategy` symbol from the `livecheck` block.
3. `LivecheckStrategy`: This is the module where strategies are defined.

The reason why we have `livecheck_strategy` is because we have conditions that should only be true when a `strategy` is defined in the `livecheck` block.

That said, I'll take another pass at this to see if I can improve this a little.

samford:
With the latest changes, `livecheck_strategy` is simply a variable corresponding to `livecheck.strategy`, which we cache to avoid executing that method more than once. This brings it inline with the `livecheck_regex` and `livecheck_url` variables, so it's easier to understand.

This also allowed me to simplify some other related code, so I think this is an improvement overall. Thanks for the feedback here.

cmd/livecheck.rb
```diff
@@ -370,9 +372,11 @@ def self.latest_version(formula)
         next
       end
 
-      url = preprocess_url(original_url)
+      # Do not process the URL when livecheck.strategy is set to :page_match
```
maxim-belkin, line 375: process ~> preprocess (?)

samford: This is perhaps a can of worms from a language perspective but I'll update it for the sake of simplicity, ahah.

cmd/livecheck.rb
```diff
@@ -384,6 +388,16 @@ def self.latest_version(formula)
         puts "Regex:            #{livecheck_regex.inspect}\n" unless livecheck_regex.nil?
       end
 
+      if livecheck_strategy == :page_match && !livecheck_regex.present?
+        odebug "#{strategy_name} strategy requires a regex"
+        next
+      end
+
+      unless livecheck_strategy.nil? || strategies.include?(strategy)
+        odebug "#{strategy_name} strategy does not apply to this URL"
```
maxim-belkin, line 397: I think Mike will ask to change this to an `if`
```rb
      if livecheck_strategy.present? && !strategies.include?(strategy)
        odebug "#{strategy_name} strategy does not apply to this URL"
```

samford: Ooo, yeah, I prefer this anyway (since I dislike `unless`). Thanks.

samford:
> ...it might help to rename `strategies` to `applicable_strategies`.

In general I agree that a more explicit name would make it clear that this variable only contains the strategies that apply to the URL (not all available strategies, as in the private `LivecheckStrategy#strategies` method). If we went this route, I would probably go for something a little less verbose like `usable_strategies`, as seen in `LivecheckStrategy#from_url`.

However, I'm not sure if this would be a wholly beneficial change. I would still want to use "strategies" in `version_info["meta"]["strategies"]` and in the debug output (i.e., `Strategies:       GithubLatest, Git`), where the purpose is clear from the context. This name mismatch may end up confusing people who are looking at the code but maybe that's not a real concern.

I'm fine with discussing the variable name change a little more here but it's something that we should handle in a separate PR, so I'm going to go ahead and merge this.

PR #1230, opened 2020-07-28
Capture Git errors in JSON output

samford:
Currently we have an issue where `brew livecheck --json` output isn't valid JSON if the run includes a Git error. Currently the `pg_top` formula produces a Git error (until/unless Homebrew/homebrew-core#56262 is merged) and it ends up being printed before the JSON:

```
fatal: repository 'https://git.postgresql.org/gitweb/?p=pg_top.git/' not found
[{"formula":"pg_top","status":"error","messages":["Unable to get versions"]}]
```

This PR is a first attempt at capturing Git errors in the JSON output. The JSON output on this branch follows the existing format for errors and contains the Git error in the `messages` array:

```json
[{"formula":"pg_top","status":"error","messages":["fatal: repository 'https://git.postgresql.org/gitweb/?p=pg_top.git/' not found"]}]
```

This implementation works fine but I'm not sure if it's quite as nice as I would like it to be, so some feedback on this would be welcome.

For what it's worth, part of the reason why I made the `Git#git_tags` changes like I did was because it may be extended in the future to include the commit hash for the tag in the JSON response. This particular feature was requested by some folks who were working on creating an automated-PR-creation tool.

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -152,6 +152,8 @@ module Homebrew
 
       if latest.nil?
         if Homebrew.args.json?
+          next version_info if version_info.is_a?(Hash) && version_info["status"] && version_info["messages"]
+
           next status_hash(formula, "error", [NO_VERSIONS_MSG])
         else
           raise TypeError, NO_VERSIONS_MSG
@@ -208,7 +210,7 @@ module Homebrew
   end
 
   def self.formula_name(formula)
-    Homebrew.args.full_name? ? formula.full_name : formula
+    Homebrew.args.full_name? ? formula.full_name : formula.name
   end
   private_class_method :formula_name
 
@@ -361,7 +363,7 @@ module Homebrew
       puts "Livecheckable?:   #{has_livecheckable ? "Yes" : "No"}"
     end
 
-    urls.each do |original_url|
+    urls.each_with_index do |original_url, i|
       puts "\nURL:              #{original_url}" if Homebrew.args.debug?
 
       # Skip Gists until/unless we create a method of identifying revisions
@@ -389,6 +391,13 @@ module Homebrew
       strategy_data = strategy.find_versions(url, livecheck_regex)
       match_version_map = strategy_data[:matches]
       regex = strategy_data[:regex]
+
+      if strategy_data[:messages].is_a?(Array) && match_version_map.empty?
+        puts strategy_data[:messages] unless Homebrew.args.json?
+        next if i + 1 < urls.length
+
+        return status_hash(formula, "error", strategy_data[:messages])
+      end
 
       if Homebrew.args.debug?
         puts "URL (strategy):   #{strategy_data[:url]}" if strategy_data[:url] != url
--- a/livecheck/livecheck_strategy/git.rb
+++ b/livecheck/livecheck_strategy/git.rb
@@ -1,17 +1,28 @@
 # frozen_string_literal: true
 
+require "open3"
+
 module LivecheckStrategy
   class Git
     PRIORITY = 8
 
     def self.tag_info(repo_url, filter = nil)
-      raw_tags = `GIT_TERMINAL_PROMPT=0 git ls-remote --tags #{repo_url}`
-      raw_tags.gsub!(%r{^.*\trefs/tags/}, "")
-      raw_tags.delete_suffix!("^{}")
+      stdout_str, stderr_str, _status = Open3.capture3(
+        { "GIT_TERMINAL_PROMPT" => "0" }, "git", "ls-remote", "--tags", repo_url
+      )
 
-      tags = raw_tags.split("\n").uniq.sort
+      tags_data = { tags: [] }
+      tags_data[:messages] = stderr_str.split("\n") unless stderr_str.empty?
+      return tags_data if stdout_str.empty?
+
+      stdout_str.gsub!(%r{^.*\trefs/tags/}, "")
+      stdout_str.delete_suffix!("^{}")
+
+      tags = stdout_str.split("\n").uniq.sort
       tags.select! { |t| t =~ filter } if filter
-      tags
+      tags_data[:tags] = tags
+
+      tags_data
     end
     private_class_method :tag_info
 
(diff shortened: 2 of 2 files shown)


livecheck/livecheck_strategy/git.rb
```diff
@@ -1,18 +1,29 @@
...
-      raw_tags = `GIT_TERMINAL_PROMPT=0 git ls-remote --tags #{repo_url}`
-      raw_tags.gsub!(%r{^.*\trefs/tags/}, "")
-      raw_tags.delete_suffix!("^{}")
+      stdout_str, stderr_str, _status = Open3.capture3(
+        { "GIT_TERMINAL_PROMPT" => "0" }, "git", "ls-remote", "--tags", repo_url
+      )
+
+      tags_data = { tags: [] }
+      tags_data[:messages] = stderr_str.split("\n") unless stderr_str.empty?
+      tags_data if stdout_str.empty?
```
maxim-belkin, line 17: did you want to `return` here?

maxim-belkin:
1. Output of `brew livecheck --json pg_top` looks better with this change than without it.
1. I think it would been nice if output of `brew livecheck pg_top` looked like this:
    ```sh
    Error: pg_top: Unable to get versions (message: "fatal: repository 'https://git.postgresql.org/gitweb/?p=pg_top.git/' not found")
    ```
1. why do you use `messages` and not `message`?

cmd/livecheck.rb
```diff
@@ -380,6 +384,16 @@ def self.latest_version(formula)
       match_version_map = strategy_data[:matches]
       regex = strategy_data[:regex]
 
+      if strategy_data[:messages].is_a?(Array) && match_version_map.empty?
+        puts strategy_data[:messages] unless Homebrew.args.json?
+
+        if i + 1 >= urls.length
```
maxim-belkin, line 390: what is the logic here? it would be nice to have a brief comment above explaining it

livecheck/livecheck_strategy/git.rb
samford: Whoops, I did mean to return here. Thanks for the catch.

cmd/livecheck.rb
samford: The general idea is that we want livecheck to be able to check other URLs after the Git error, like it currently does. If we simply `return` whenever there's an error within a strategy, this would prevent livecheck from continuing on to the other URLs. As such, we only return the error when there aren't other URLs to check.

Does this make the code clear or do you think it could still use a comment? If you still feel it's necessary, what do you think about `# Allow livecheck to try other URLs, when available`?

samford:
> I think it would've been nice if output of `brew livecheck pg_top` looked like this:
>   ```shell
>   Error: pg_top: Unable to get versions (message: "fatal: repository 'https://git.postgresql.org/gitweb/?p=pg_top.git/' not found")
>   ```

This PR is about the JSON output, so I've simply maintained the existing behavior for the default and debug output. This is something that we can discuss in an issue or a separate PR but I'm generally fine with the existing behavior.

> why do you use `messages` and not `message`?

Currently there's only one message in the `messages` array but there could theoretically be more at some point. I would only name it `message` if the value is a string. This is something that we can discuss in the future when we talk about the JSON format in general (sometime after the migrations are done).

cmd/livecheck.rb
maxim-belkin: Makes sense but I'd suggest using something like `original_url.equal? urls.last`

samford: `original_url.equal? urls.last` compares the values (instead of the index), so it's not functionally equivalent and doesn't represent the same intention here. This would introduce the potential for this condition to act inappropriately under unusual circumstances (i.e., a URL array like [A, B, A] would prevent B from being checked), so it doesn't seem compelling to me.

maxim-belkin: Can you limit it to `==`?

samford: `>=` is just defensive programming in this context, as it matches the correct situation where `i + 1 == urls.length` but will continue to work if `i + 1` is ever greater than `urls.length`. It _shouldn't_ ever be greater than `urls.length` but it handles that error condition gracefully, whereas `==` wouldn't. There's no detriment to using `>=` here whereas there is a potential benefit.

samford: I flipped the condition to `i + 1 < urls.length`, which accomplishes the same thing but may resolve your concern.

cmd/livecheck.rb
```diff
@@ -390,6 +394,16 @@ def self.latest_version(formula)
...
       regex = strategy_data[:regex]
 
+      if strategy_data[:messages].is_a?(Array) && match_version_map.empty?
+        puts strategy_data[:messages] unless Homebrew.args.json?
+
+        if i + 1 < urls.length
+          next
+        else
+          return status_hash(formula, "error", strategy_data[:messages])
+        end
```
maxim-belkin, line 404:
```suggestion
        next if i + 1 < urls.length
        return status_hash(formula, "error", strategy_data[:messages])
```

(feel free to modify your commit and force-push)

samford: I thought about doing this but `brew style` will want an empty line after the guard clause, so it ends up being:

```ruby
next if i + 1 < urls.length

return status_hash(formula, "error", strategy_data[:messages])
```

I figured this wasn't that far off from the existing if/else and was maybe less clear. However, if we move the guard underneath the `puts`, I think it's fine:

```ruby
puts strategy_data[:messages] unless Homebrew.args.json?
next if i + 1 < urls.length

return status_hash(formula, "error", strategy_data[:messages])
```

I'll go ahead and integrate this, as it is a bit tidier.

cmd/livecheck.rb
```diff
@@ -152,6 +152,10 @@ def livecheck
 
       if latest.nil?
         if Homebrew.args.json?
+          unless version_info.nil? || !(version_info.key?("status") && version_info.key?("messages"))
+            next version_info
```
maxim-belkin, line 156: I think these two lines can be split into two `next` llines

maxim-belkin: something like
```suggestion
          next version_info if !version_info.nil?
          next version_info unless version_info.key?("status") && version_info.key?("messages")
```

samford: `brew style` is going to want `if !version_info.nil?` to be `unless version_info.nil?`, so you end up with the following:

```ruby
next version_info unless version_info.nil?
next version_info unless version_info.key?("status") && version_info.key?("messages")
```

This strikes me as more repetitious than the existing setup. The only reason why have to use `unless/end` here instead of the modifier form of `unless` is that the line is one character beyond our 118 character limit.

How about:

```ruby
next version_info if version_info.is_a?(Hash) && version_info["status"] && version_info["messages"]
```

`version_info["status"]` and `version_info["messages"]` will give `nil` if the corresponding key isn't present, so it's roughly the same thing.

samford:
@maxim-belkin How do you feel about the current state of this? Our timeline for the Homebrew/brew migration has shifted and will now start at the end of this week or earlier, so I'm trying to merge this before the end of the day.

PR #1231, opened 2020-07-28
autossh: update regex

nandahkrishna:
This PR brings `autossh`'s regex up to current standards. I didn't want to deviate from the earlier regex too much so I've used `[a-z]+`, just as before.

### DIFF
--- a/Livecheckables/autossh.rb
+++ b/Livecheckables/autossh.rb
@@ -1,6 +1,6 @@
 class Autossh
   livecheck do
     url :homepage
-    regex(/href=.*?autossh[._-]v?(\d+(?:\.\d+)+[a-z]+)\.t/i)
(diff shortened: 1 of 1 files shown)


Livecheckables/autossh.rb
```diff
@@ -1,6 +1,6 @@
 class Autossh
   livecheck do
     url :homepage
-    regex(/href=.*?autossh-([0-9.]+[a-z]+)\.t/i)
+    regex(/href=.*?autossh[._-]v?(\d+(?:\.\d+)+[a-z]+)\.t/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?autossh[._-]v?(\d+(?:\.\d+)+[a-z]?)\.t/i)
```

Looking at the [changelog](https://www.harding.motd.ca/autossh/CHANGES.txt), the trailing letter isn't present on the first release (i.e., `1.2` comes before `1.2a`), so we need to make it optional and I think we should only match one character.

PR #1232, opened 2020-07-29
Update regexes containing `a-z` and `0-9`

nandahkrishna:
This PR updates regexes containing `0-9` and `a-z`.

Attention required:
* `cscope`: We were matching a version ending in `bl2`, hence the `*`
* `isc-dhcp`: Seems like it doesn't need `[a-z]`
* `mksh`: Made `R` optional (`R?` just like `v?`)
* `x264`: No change in capture group

### DIFF
--- a/Livecheckables/isc-dhcp.rb
+++ b/Livecheckables/isc-dhcp.rb
@@ -1,6 +1,6 @@
 class IscDhcp
   livecheck do
     url "https://www.isc.org/downloads/"
-    regex(%r{/dhcp-([a-zA-Z0-9.]+)\.t}i)
+    regex(%r{/dhcp-v?(\d+(?:\.\d+)+)\.t}i)
   end
 end
--- a/Livecheckables/isc-dhcp.rb
+++ b/Livecheckables/isc-dhcp.rb
@@ -1,6 +1,6 @@
 class IscDhcp
   livecheck do
     url "https://www.isc.org/downloads/"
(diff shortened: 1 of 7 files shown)


nandahkrishna:
The livecheck for `discount` seems to work just fine for me locally, but CI failed because of a `certificate verify failed (unable to get local issuer certificate)` error.

Livecheckables/isc-dhcp.rb
```diff
@@ -1,6 +1,6 @@
 class IscDhcp
   livecheck do
     url "https://www.isc.org/downloads/"
-    regex(%r{/dhcp-([a-zA-Z0-9.]+)\.t}i)
+    regex(%r{/dhcp-v?(\d+(?:\.\d+)+)\.t}i)
```
samford, line 4:
```suggestion
    regex(%r{/dhcp-v?(\d+(?:\.\d+)+(?:-P\d+)?)\.t}i)
```

Looking at the [release notes](https://gitlab.isc.org/isc-projects/dhcp/-/blob/master/RELNOTES), the releases with letters are typically alpha, beta, and release candidates, so I'm good with omitting those.

However, ISC DHCP does sometimes have patch releases like `4.3.6-P1` and these have been used in the formula in the past, so I think we should continue to match those.

samford:
I switched CI back to `macOS-latest` and everything works again, as expected.

PR #1236, opened 2020-07-29
vsts-cli: remove livecheckable

nandahkrishna:
The Livecheckable was unnecessary thanks to the PyPI strategy. Also, I found this on the "homepage" linked to on PyPI (which happens to be this [repo](https://github.com/Azure/azure-devops-cli-extension)):
```
The Azure CLI with the Azure DevOps Extension has replaced the VSTS CLI. The VSTS CLI
has been deprecated and will no longer be receiving new features. We recommend that users
of the VSTS CLI switch to the Azure CLI and add the Azure DevOps extension.
```

### DIFF
--- a/Livecheckables/vsts-cli.rb
+++ b/Livecheckables/vsts-cli.rb
@@ -1,6 +1,0 @@
-class VstsCli
-  livecheck do
-    url :stable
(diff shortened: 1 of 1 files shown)


nandahkrishna:
I've created a PR in homebrew-core (Hombrew/homebrew-core#58760) to deprecate the Formula.

PR #1238, opened 2020-07-29
wwwoffle: update url and regex

samford:
This brings the existing `wwwoffle` livecheckable up to current standards:

* Align the check with the location of the stable archive, when possible
* Use `href=.*?` to match the file name within the `href` attribute, when applicable
* Use the `[._-]` delimiter between the software name and version
* Use `v?(\d+(?:\.\d+)+[a-z]?)` instead of `([a-z0-9.]+)`, to be more clear about what we want to match

### DIFF
--- a/Livecheckables/wwwoffle.rb
+++ b/Livecheckables/wwwoffle.rb
@@ -1,6 +1,6 @@
 class Wwwoffle
   livecheck do
-    url :homepage
-    regex(/The absolute latest version is version ([a-z0-9.]+)\./i)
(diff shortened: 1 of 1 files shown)


nandahkrishna:
Thanks! I'd just pushed my changes for this one to a branch (forgot to do so yesterday), and they were identical.

samford:
I saw your branch was pushed a minute after I created this PR and I wasn't at all surprised that it was identical, ahah.

PR #1244, opened 2020-07-30
Update delimiter before `v?` to [._-]

nandahkrishna:
This PR updates delimiters between the Formula name and `v?` in the regex to `[._-]` from the existing `.` or `-` or `_`. The matches output was the same before and after when I checked, and there were no errors.


nandahkrishna:
Ah it seems I missed that while going through the colordiff output, should probably just use a script instead.

PR #1246, opened 2020-07-31
Update delimiter before capturing group to [._-] and add v?

nandahkrishna:
This PR updates delimiters before the capturing group to `[._-]` and also adds `v?`.

I compared the matched versions before and after with this script, where `before` and `after` are the JSON livecheck output files read using `JSON.parse`, and list contains all the Livecheckables I anticipated would change.

```ruby
for i in 0..list.length do
  if before[i]["matches"] != after[i]["matches"]
    puts "Hi #{i}"
  end
end
```

There seemed to be no difference though.

### DIFF
--- a/Livecheckables/a52dec.rb
+++ b/Livecheckables/a52dec.rb
@@ -1,6 +1,6 @@
 class A52dec
   livecheck do
     url "http://liba52.sourceforge.net/downloads.html"
-    regex(/href=.*?a52dec-(\d+(?:\.\d+)+)\.t/i)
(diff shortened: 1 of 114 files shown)


nandahkrishna:
It seems `atool` failed CI, just checked and it says `404 Not Found` for me too.

samford:
> I've updated the Livecheckable to use this url and also forced it to use `strategy :page_match`. I've made the change locally, should I add it to this PR

This PR needed to be rebased (to get some fixes from the latest main branch), so I went ahead and added an update for `atool` in the process. Let me know if this differs from your implementation.

nandahkrishna:
Nope, it's the same as what I had in mind. 😀

PR #1248, opened 2020-07-31
Fix references to Homebrew.args

SeekingMeaning:
Homebrew/brew#8144

`livecheck/livecheck_strategy/gnu.rb` and `livecheck/extend/formulary.rb` might also need to be updated

Just thought I would open this to provide a starting point

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -76,27 +76,27 @@ module Homebrew
   end
 
   def livecheck
-    livecheck_args.parse
+    @args = livecheck_args.parse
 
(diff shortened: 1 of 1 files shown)


cmd/livecheck.rb
```diff
@@ -76,27 +76,27 @@ def livecheck_args
   end
 
   def livecheck
-    livecheck_args.parse
+    @args = livecheck_args.parse
```
reitermarkus, line 79: Pass this to other methods rather than making it an instance variable.

PR #1249, opened 2020-07-31
Update CLI::Parser args handling

samford:
This updates all the current usage of `Homebrew.args` to use an `args` variable instead. In line with how the Homebrew/brew commands handle this, I've updated the functions to accept a named `args` argument.

The instance in the `Gnu` strategy has simply been replaced with `odebug`, as that's more appropriate.

Closes #1248 (I discussed this with Seeker and they were fine with me opening this PR that closes theirs).

### DIFF
--- a/cmd/livecheck.rb
+++ b/cmd/livecheck.rb
@@ -76,27 +76,27 @@ module Homebrew
   end
 
   def livecheck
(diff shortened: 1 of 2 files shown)


reitermarkus:
Thanks!

PR #1277, opened 2020-08-14
GitHub Actions cleanup

MikeMcQuaid:
- Remove `pull_request: []` warning
- Use more Homebrew actions
- Use implicit `run` `name`s

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,31 +3,19 @@ name: GitHub Actions CI
 on:
   push:
     branches: master
-  pull_request: []
+  pull_request:
 
 jobs:
   tests:
     runs-on: macOS-latest
-    env:
-      HOMEBREW_NO_AUTO_UPDATE: 1
-      HOMEBREW_NO_ANALYTICS: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -3,34 +3,19 @@ name: GitHub Actions CI
...
-        ln -s $PWD $(brew --repo $GITHUB_REPOSITORY)
-
-    - name: Run brew style
-      run: brew style $GITHUB_REPOSITORY
-
-    - name: Run brew livecheck
-      run: brew livecheck
-
-    - name: Fetch origin/master
-      run: git fetch --depth=1 origin master
```
dawidd6, line 33: I think this step should be left as is

MikeMcQuaid: We're already fetching the tap `Homebrew/actions/setup-homebrew@master` no? Is the issue that we're not fetching `origin/master` specifically?

dawidd6: Yep, it's needed in later step

PR #1289, opened 2020-08-15
go@1.14: update url and regex

chenrui333:
relates to the recent go1.15 upgrade.

```
$ brew livecheck go@1.14
go@1.14 : 1.14.7 ==> 1.14.7
```

### DIFF
--- a/Livecheckables/go@1.14.rb
+++ b/Livecheckables/go@1.14.rb
@@ -1,0 +1,6 @@
+class GoAT114
+  livecheck do
+    url "https://golang.org/dl/"
+    regex(/href=.*?gov?(1\.14(?:\.\d+)+)\.src/i)
+  end
+end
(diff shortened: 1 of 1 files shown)


Livecheckables/go@1.14.rb
```diff
@@ -0,0 +1,6 @@
+class GoAT114
+  livecheck do
+    url "https://golang.org/dl/"
+    regex(/href=.*?gov?(1\.14(?:\.\d+)+)\.src/i)
```
samford, line 4:
```suggestion
    regex(/href=.*?gov?(1\.14(?:\.\d+)*)\.src/i)
```

The first Go release for a given minor release only includes the major and minor (no patch). Since this regex specifies the major and minor version, we should make the patch optional (i.e., `(?:\.\d+)*` instead of `(?:\.\d+)+`).

chenrui333: > I've also made some other changes to bring this more in line with our current standards.  I'll create a PR to update the `go` livecheckable in the same fashion.

👍

PR #1290, opened 2020-08-15
helm@2: add livecheckable

chenrui333:
helm 2.x is still actively maintained, it is good to keep it under the watch. The most recent release was three days back.

### DIFF
--- a/Livecheckables/helm@2.rb
+++ b/Livecheckables/helm@2.rb
@@ -1,6 +1,6 @@
 class HelmAT2
   livecheck do
-    url "https://github.com/helm/helm.git"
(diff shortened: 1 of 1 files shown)


Livecheckables/helm@2.rb
```diff
@@ -0,0 +1,6 @@
+class HelmAT2
+  livecheck do
+    url "https://github.com/helm/helm.git"
```
samford, line 3:
```suggestion
    url :stable
```

chenrui333: Good point!

PR #1329, opened 2020-08-25
Refactor and document strategies

samford:
The primary goals of this PR are to refactor the regexes in strategies to make them more comprehensible and to add documentation to explain how the strategies work in general.

Some of the changes here are as follows:

* Make the default priority for strategies a constant and explain the priority setup.
* Split complex regexes onto multiple lines and add explanatory comments, to help make them more understandable.
* Use named capture groups in regexes, where possible.
* Use `<<~EOS`/`EOS` instead of newlines in string literals.
* Uniformly use `Regexp#escape` when interpolating variables in strategy default regexes.

One thing to note is that I didn't split the default regex literals onto multiple lines. These regexes are printed in debug output and included in verbose JSON output, so multiline regexes (with comments) would be a mess in those contexts.

I'm sure there are areas for improvement here but I think this is as much as I can do before review.

### DIFF
--- a/livecheck/strategy.rb
+++ b/livecheck/strategy.rb
@@ -1,8 +1,26 @@
 # frozen_string_literal: true
 
+# The `Livecheck::Strategy` module contains the various strategies as well
+# as some general-purpose methods for working with them. Within the context
+# of the `brew livecheck` command, strategies are established procedures
+# for finding new software versions at a given source.
 module LivecheckStrategy
   module_function
 
+  # Strategy priorities informally range from 1 to 10, where 10 is the
+  # highest priority. 5 is the default priority because it's roughly in
+  # the middle of this range. Strategies with a priority of 0 (or lower)
+  # are ignored.
+  DEFAULT_PRIORITY = 5
+  private_constant :DEFAULT_PRIORITY
+
+  # Creates and/or returns a `@strategies` `Hash` ,which maps a snake
+  # case strategy name symbol (e.g., `:page_match`) to the associated
+  # `Strategy`.
+  #
+  # At present, this should only be called after tap strategies have been
+  # loaded, otherwise livecheck won't be able to use them.
+  # @return [Hash]
   def strategies
     return @strategies if defined? @strategies
 
@@ -16,12 +34,23 @@ module LivecheckStrategy
   end
   private_class_method :strategies
 
+  # Returns the `Strategy` that corresponds to the provided `Symbol` (or
+  # `nil` if there is no matching `Strategy`).
+  # @param symbol [Symbol] the strategy name in snake case as a `Symbol`
+  #   (e.g., `:page_match`)
+  # @return [Strategy, nil]
   def from_symbol(symbol)
     strategies[symbol]
   end
 
+  # Returns an array of strategies that apply to the provided URL.
+  # @param url [String] the URL to check for matching strategies
+  # @param regex_provided [Boolean] whether a regex is provided in a
+  #   `livecheck` block
+  # @return [Array]
   def from_url(url, regex_provided = nil)
     usable_strategies = strategies.values.select do |strategy|
+      # Ignore strategies with a priority of 0 or lower
       next if strategy.const_defined?(:PRIORITY) && !strategy::PRIORITY.positive?
 
       strategy.respond_to?(:match?) && strategy.match?(url)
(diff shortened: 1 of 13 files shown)


livecheck/strategy.rb
```diff
@@ -1,8 +1,26 @@
...
+# for finding new software versions at a given source.
 module LivecheckStrategy
   module_function
 
+  # Strategy priorities informally range from 1 to 10, where 10 is the
+  # highest priority. 5 is the default priority because it's roughly in
+  # the middle of this range. Strategies with a priority of 0 (or lower)
+  # are ignored.
+  DEFAULT_PRIORITY = 5
+  private_constant :DEFAULT_PRIORITY
```
samford, line 15: This has to be private right now because we rely on strategies being the only constants present in the `Strategy` module. If this isn't private, then livecheck treats it like a strategy and runs into an error.

We can avoid this in the future if we we rework how we create the `@strategies` hash but I think we should defer that until after GSoC, due to time constraints.

livecheck/strategy.rb
```diff
@@ -16,23 +34,34 @@ def strategies
   end
   private_class_method :strategies
```
samford, line 35: I left this private for now because I feel like it would take too much time to rework how we're creating the `@strategies` hash right now. As mentioned above, this is something that I think we should work on after GSoC, when we're not staring down a deadline.

livecheck/strategy/git.rb
```diff
@@ -44,15 +81,15 @@ def find_versions(url, regex = nil)
...
-        # Skip tag if it has a 'debian/' prefix and upstream does not do only
-        # 'debian/' prefixed tags
+        # Skip tag if it has a 'debian/' prefix and upstream does not do
+        # only 'debian/' prefixed tags
         next if tag =~ %r{^debian/} && !tags_only_debian
 
         captures = regex.is_a?(Regexp) ? tag.scan(regex) : []
         tag_cleaned = if captures[0].is_a?(Array)
-          captures[0][0] # Extract the first capture group
+          captures[0][0] # Use the first capture group (the version)
```
samford, line 90: Just to be clear, we can't use a named capture group here because the regexes are coming from `livecheck` blocks. We would have to modify the regexes in tons of `livecheck` blocks to use a named capture group for this to work.

livecheck/strategy/apache.rb
```diff
@@ -1,21 +1,64 @@
...
 
+    # Generates a URL and regex (if one isn't provided) and passes them
+    # to the `PageMatch#find_versions` method to identify versions in the
+    # content.
+    # @param url [String] the URL of the content to check
+    # @param regex [Regexp] a regex used for matching versions in content
+    # @return [Hash]
     def find_versions(url, regex = nil)
-      path, prefix, suffix = url.match(%r{path=(.+?)/([^/]*?)\d+(?:\.\d+)+(/|[^/]*)}i)[1, 3]
+      m = url.match(URL_PART_REGEX).named_captures
```
MikeMcQuaid, line 49:
```suggestion
      captures = url.match(URL_PART_REGEX).named_captures
```

and same elsewhere

MikeMcQuaid (approved):
Looks good to me! One tiny nit but not even that is really blocking. Thanks for the cleanup @samford!

livecheck/strategy/apache.rb
samford: Though I'm not fond of non-descriptive variables, I was attempting to minimize the verbosity of `captures["prefix"]` vs `prefix`. Ideally it would just be `URL_PART_REGEX =~ url` but only regex literals create local variables from named capture groups when using `=~`.

My initial thinking around using a `URL_PART_REGEX` constant was that we wouldn't need to create a regex literal on every execution of `#find_versions`. However, after profiling this, using a regex literal in `#find_versions` with `=~` actually uses less memory (no hash key allocations) and is ever-so-slightly faster.

I'm going to convert the related regex constants into literals that use `=~`, as we'll simply have `path`, `prefix`, `suffix`, etc. local variables without dealing with the verbosity of the `#named_captures` hash setup.

MikeMcQuaid: My main thinking here is: this is not a match but is the named captures. 



> I'm going to convert the related regex constants into literals that use `=~`, as we'll simply have `path`, `prefix`, `suffix`, etc. local variables without dealing with the verbosity of the `#named_captures` hash setup.

sounds better 👍🏻

samford:
I went through this one more time and nothing really sticks out at this point, so I'm going to go ahead and merge this. Thanks for the review!

